PR #70, opened 2015-01-07
Change from AGPL to Apache

nedbat:
@clytwynec OK? I can explain if you like.


NOTICE
```diff
@@ -0,0 +1,3 @@
+bok_choy
+
+Copyright 2013-2014, edX
```
clytwynec: Should this be 2015?  (I really don't know, but thought I'd ask just in case.)

jzoldak:
Regarding the travis failure - filed as Issue #71

wedaly:
Merged the fix, so tests should pass if this is rebased onto master.

PR #74, opened 2015-01-12
Update psutil to 2.2.1

singingwolfboy:


cgoldberg:
lgtm.
was this previously broken? (since get_children was deprecated in 2.0)
:+1:

singingwolfboy:
@cgoldberg Looks like `get_children` should work, but raises a DeprecationWarning, according to https://github.com/giampaolo/psutil/blob/master/HISTORY.rst

PR #76, opened 2015-01-13
add Promise for browser start

clytwynec:
@jzoldak  Please review.  This should catch the error described in [TE-701](https://openedx.atlassian.net/browse/TE-701), and will just retry starting the browser.


benpatterson:
:+1:

PR #65, opened 2014-09-24
Updated to Selenium 2.43.0

clintonb:
Fixes #64 

Do we trust Selenium enough to not break their API and do >= and avoid having to continuously update this file?

@clytwynec @jzoldak

PR #77, opened 2015-01-14
Bypass media device access prompt

rlucioni:
Uses a Firefox profile to bypass the security prompt displayed when the browser attempts to access a media device (e.g., a webcam). Also updates the AUTHORS file.

@wedaly, could you take a look at this when you're able?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -9,3 +9,4 @@ Dennis Jen <djen@edx.org>
 Muhammad Ammar <mammar@gmail.com>
 Ben Patterson <bpatterson@edx.org>
 Minh-Tue Vo <minhtuev@gmail.com>
+Renzo Lucioni <renzo@edx.org>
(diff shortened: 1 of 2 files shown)


wedaly:
:+1:  When you update the requirements in edx-platform, please pin "selenium==2.42.1" (per discussion here: https://github.com/edx/bok-choy/pull/65)

rlucioni:
Thanks for the review. I might be mistaken, but it looks like the selenium version used by edx-platform was pinned to 2.42.1 by [this commit](https://github.com/edx/edx-platform/commit/688909eb76d503931838a40069bfd5a571fd3306#diff-f9c61f9375a7917349c8217216293eb1R127). I'm planning on updating [github.txt](https://github.com/edx/edx-platform/blob/master/requirements/edx/github.txt#L29) so that these changes to bok-choy are pulled in.

PR #78, opened 2015-01-15
Add support for faking media device access in Chrome

rlucioni:
@jzoldak, as requested. (Reference on the Chrome webdriver's Options class [here](https://code.google.com/p/selenium/source/browse/py/selenium/webdriver/chrome/options.py).)

@wedaly, FYI.


rlucioni:
@wedaly would it be possible for you to take a quick look at this, when you're able? @jzoldak mentioned he won't have an opportunity to look at it today.

wedaly:
:+1:

PR #80, opened 2015-03-13
Move browser config setup into try block TE-783

jzoldak:
Follow on to #76. 
See https://openedx.atlassian.net/browse/TE-783
Error message given is: `Errno 17 File exists: '/tmp/<tmpdir>/extensions/fxdriver@googlecode.com'`

When no FF profile is specified, the FirefoxProfile __init__ method uses tempfile.mkdtemp() to create a new temp dir. This part also needs to be re-executed in the case that the browser did not start up successfully.

@clytwynec

### DIFF
--- a/bok_choy/browser.py
+++ b/bok_choy/browser.py
@@ -168,27 +168,28 @@ def browser(tags=None, proxy=None):
 
     browser_name = os.environ.get('SELENIUM_BROWSER', 'firefox')
 
-    # Get the class and kwargs required to instantiate the browser based on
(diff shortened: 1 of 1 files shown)


clytwynec:
Seems good to me. Will we be able to confirm that this fixes the issue now?

:+1:

jzoldak:
@clytwynec only by the absence of issues. (I guess we could have printed something out before the 2nd try so we could tell via the console when multiple attempts were required to start up the browser successfully.)

PR #82, opened 2015-03-25
Fix for issue in js_defined (TNL-1770)

clytwynec:
@benpatterson

### DIFF
--- a/tests/pages.py
+++ b/tests/pages.py
@@ -255,6 +255,23 @@ class JavaScriptPage(SitePage):
         self.q(css='div#fixture button').first.click()
 
 
+@js_defined('something.SomethingThatDoesntExist')
+class JavaScriptUndefinedPage(SitePage):
+    """
+    Page for testing asynchronous JavaScript, where the
+    javascript that we wait for is never defined.
+    """
+
+    name = "javascript"
+
+    @wait_for_js
+    def trigger_output(self):
+        """
+        Click a button which will only work once RequireJS finishes loading.
+        """
+        self.q(css='div#fixture button').first.click()
+
+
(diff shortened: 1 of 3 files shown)


tests/test_javascript.py
```diff
@@ -23,6 +24,12 @@ def test_wait_for_defined_with_reload(self):
         javascript.reload_and_trigger_output()
         self.assertEquals(javascript.output, "Done")
 
+    def test_wait_for_defined_failure(self):
+        with self.assertRaises(BrokenPromise):
```
clytwynec: Without the above changes in javascript.py,  this test raises:
```
WebDriverException: Message: something is not defined
```

With the changes, it will correctly keep trying and timeout, raising a BrokenPromise.

tests/pages.py
```diff
@@ -255,6 +255,23 @@ def reload_and_trigger_output(self):
         self.q(css='div#fixture button').first.click()
 
 
+@js_defined('something.SomethingThatDoesntExist')
+class JavaScriptUndefinedPage(SitePage):
+    """
+    Page for testing asynchronous JavaScript, where the
+    javascript what we wait for is never defined.
```
benpatterson: nit: s/what/that

tests/test_javascript.py
benpatterson: Makes sense. thanks for pointing that out.

benpatterson:
Assuming the failed build is unrelated, :+1:

clytwynec:
It looks like the tutorial is broken again because of some changing identifiers in the github UI.  I opened PR #83 to address that failure.

PR #85, opened 2015-04-07
Add ability to save browser page source to disk

jzoldak:
@cgoldberg @mlkwaqas

### DIFF
--- a/bok_choy/browser.py
+++ b/bok_choy/browser.py
@@ -2,9 +2,9 @@
 Use environment variables to configure Selenium remote WebDriver.
 For use with SauceLabs (via SauceConnect) or local browsers.
 """
+import os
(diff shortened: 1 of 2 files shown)


jzoldak:
I think all that is left for #86 is to save the page source on failure.

dianakhuang:
@jzoldak @benpatterson Could you review this PR? I believe I handled the merge conflicts correctly, but I'd like some more eyes on it.

Also, the test failure is related to HAR capture, which is a feature we're removing anyway.

dianakhuang:
@jmbowman @ajpal @thallada

dianakhuang:
We've got a clean build, hooray!

jzoldak:
👍

PR #87, opened 2015-04-30
TE-817 Enable accessibility auditing

jzoldak:
using PhantomJS and Google Chrome Accessibility Developer Tools
@benpatterson @cgoldberg @cptvitamin

### DIFF
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -2,19 +2,26 @@
 Base implementation of the Page Object pattern.
 See https://code.google.com/p/selenium/wiki/PageObjects
 """
-
-import logging
-import socket
-import urlparse
 from abc import ABCMeta, abstractmethod, abstractproperty
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from functools import wraps
 from contextlib import contextmanager
+import json
+import logging
+import os
+import socket
 from textwrap import dedent
+import urlparse
+
+import requests
 from selenium.common.exceptions import WebDriverException
 
 from .query import BrowserQuery
 from .promise import Promise, EmptyPromise, BrokenPromise
+
+
+AXS_FILE = 'vendor/google/axs_testing.js'
+AuditResults = namedtuple('AuditResults', 'errors, warnings')
 
 
 class WrongPageError(Exception):
@@ -27,6 +34,13 @@ class WrongPageError(Exception):
 class PageLoadError(Exception):
     """
     An error occurred while loading the page.
+    """
+    pass
+
+
+class AccessibilityError(Exception):
+    """
+    The page violates one or more accessibility rules.
     """
     pass
 
@@ -173,6 +187,8 @@ class PageObject(object):
             PageObject
         """
         self.browser = browser
+        flag = os.environ.get('VERIFY_ACCESSIBILITY', 'False')
+        self.verify_accessibility = flag.lower() == 'true'
 
     @abstractmethod
     def is_browser_on_page(self):
@@ -321,10 +337,15 @@ class PageObject(object):
         Raises:
             BrokenPromise: The timeout is exceeded without the page loading successfully.
         """
-        return Promise(
+        result = Promise(
             lambda: (self.is_browser_on_page(), self), "loaded page {!r}".format(self),
             timeout=timeout
         ).fulfill()
+
+        if self.verify_accessibility:
+            self._check_for_accessibility_errors()
+
+        return result
 
     @unguarded
     def q(self, **kwargs):  # pylint: disable=invalid-name
@@ -500,3 +521,132 @@ class PageObject(object):
 
         """
         self.wait_for(lambda: not self.q(css=element_selector).visible, description=description, timeout=timeout)
+
+    def axs_audit_rules_to_run(self):
+        """
+        List of rules to check for accessibility errors on the page.
+        See https://github.com/GoogleChrome/accessibility-developer-tools/tree/master/src/audits
+
+        E.g. return ['badAriaAttributeValue']
+        An empty list means to check for all available rules.
+        None means that no audit should be done for this page.
+        """
+        return []
+
+    def axs_scope(self):
+        """
+        The "start point" for the audit: the element which contains the portion of
+        the page which should be audited.
+
+        E.g. return 'document.querySelector("div#foo")'
+        Defaults to using the document as the scope.
+        """
+        return 'null'
+
+    def do_axs_audit(self):
+        """
+        Use Google's Accessibility Developer Tools to audit the
+        page for accessibility problems.
+
+        See https://github.com/GoogleChrome/accessibility-developer-tools
+
+        Since this needs to inject JavaScript into the browser page, the only
+        known way to do this is to use PhantomJS as your browser.
+
+        Raises:
+            NotImplementedError if you are not using PhantomJS
+            RuntimeError if there was a problem getting the report
+
+        Returns:
+            A list (one for each browser session) of namedtuples with 'errors' and 'warnings'
+            fields whose values are the errors and warnings returned from the audit.
+
+            None if the page object has no rules defined to check.
+        """
+        if self.browser.name != 'phantomjs':
+            msg = 'Accessibility auditing is only supported with PhantomJS as the browser.'
+            raise NotImplementedError(msg)
+
+        rules = self.axs_audit_rules_to_run()
+        if rules is None:
+            msg = 'No accessibility rules were specified to check for this page: {}'.format(self)
+            self.warning(msg)
+            return None
+
+        # The ghostdriver URL will be something like this: 'http://localhost:33225/wd/hub'
+        ghostdriver_url = self.browser.service.service_url
+
+        # Get the session_id from ghostdriver so that we can inject JS into the page.
+        resp = requests.get('{}/sessions'.format(ghostdriver_url))
+        sessions = resp.json()
+
+        # report is the list that is returned, with one item for each browser session
+        report = []
+
+        for session in sessions.get('value'):
+            session_id = session.get('id')
+
+            # This line will only be included in the script if rules to check on this page
+            # are specified, as the default behavior of the js is to run all rules.
+            if len(rules) > 0:
+                rules_config = "auditConfig.auditRulesToRun = {rules};".format(
+                    rules=rules)
+            else:
+                rules_config = ""
+
+            script = dedent("""
+                var page = this;
+                page.injectJs("{file}");
+                var report = page.evaluate(function() {{
+                  var auditConfig = new axs.AuditConfiguration();
+                  {rules_config}
+                  auditConfig.scope = {scope};
+                  var run_results = axs.Audit.run(auditConfig);
+                  var audit_results = axs.Audit.auditResults(run_results)
+                  return audit_results;
+                }});
+                return report;
+            """.format(file=AXS_FILE, rules_config=rules_config, scope=self.axs_scope()))
+
+            payload = {"script": script, "args": []}
+            resp = requests.post('{}/session/{}/phantom/execute'.format(
+                ghostdriver_url, session_id), data=json.dumps(payload))
+
+            result = resp.json().get('value')
+            if result is None:
+                msg = '{} {}'.format(
+                    'No results were returned by the audit report.',
+                    'Perhaps there was a problem with the rules or scope defined for this page.')
+                raise RuntimeError(msg)
+
+            # audit_results is report of accessibility errors for that session
+            audit_results = AuditResults(errors=result.get('errors_'), warnings=result.get('warnings_'))
+            report.append(audit_results)
+
+        return report
+
+    def _check_for_accessibility_errors(self):
+        """
+        Parse the results of an axs_audit and raise a single exception
+        if there are violations.
+
+        Note that an error is only raised on errors, not on warnings.
+
+        Returns:
+            None
+
+        Raises:
+            AccessibilityError
+        """
+        errors = []
+        audit = self.do_axs_audit()
+        for session_result in audit:
+            if session_result:
+                if len(session_result.errors) > 0:
+                    errors.extend(session_result.errors)
+
+        num_errors = len(errors)
+
+        if num_errors > 0:
+            msg = "URL '{}' has {} errors: {}".format(self.url, num_errors, ", ".join(errors))
+            raise AccessibilityError(msg)
--- a/bok_choy/web_app_test.py
+++ b/bok_choy/web_app_test.py
@@ -1,16 +1,17 @@
 """
 Base class for testing a web application.
 """
-import sys
-from unittest import TestCase, SkipTest
 from abc import ABCMeta
-from uuid import uuid4
 import functools
 import os
+import sys
+from unittest import TestCase, SkipTest
+from uuid import uuid4
+from selenium.webdriver.support.events import EventFiringWebDriver
+
 from .browser import browser, save_screenshot, save_driver_logs
 from .proxy import bmp_proxy, stop_server
 from .performance import HarListener, HarCapturer
-from selenium.webdriver.support.events import EventFiringWebDriver
 
 
 class WebAppTest(TestCase):
--- a/docs/accessibility.rst
+++ b/docs/accessibility.rst
@@ -1,0 +1,189 @@
+Accessibility Testing
+======================
+
+The bok-choy framework includes the ability to test web pages for accessibility
+using `Google Accessibility Developer Tools`_.
+
+The audit rules that are checked and optionally the scope of the audit
+within the webpage itself are defined on each Page Object's definition.
+
+The general methodology for enabling accessibility auditing is:
+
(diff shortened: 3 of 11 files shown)


bok_choy/web_app_test.py
```diff
@@ -1,16 +1,23 @@
...
+
+import requests
+from selenium.webdriver.support.events import EventFiringWebDriver
+
 from .browser import browser, save_screenshot, save_driver_logs
 from .proxy import bmp_proxy, stop_server
 from .performance import HarListener, HarCapturer
-from selenium.webdriver.support.events import EventFiringWebDriver
+
+AXS_FILE = 'vendor/axs_testing.js'
```
benpatterson: What's this file for? If I were brand new to bok-choy, where could I get it?

bok_choy/web_app_test.py
```diff
@@ -86,6 +93,63 @@ def unique_id(self):
...
+    @property
+    def axs_audit(self):
+        """
+        Use Google's Accessibility Developer Tools to audit the
+        page for accessibility problems.
+
+        See https://github.com/GoogleChrome/accessibility-developer-tools
+
+        Since this needs to inject JavaScript into the browser page, the only
+        way I know how to do this is if you are using PhantomJS as your browser.
```
benpatterson: nit: "the only known way" to do this.

bok_choy/web_app_test.py
```diff
@@ -86,6 +93,63 @@ def unique_id(self):
...
+              /** @type {Array.<Element>} */
+              elements,  // The elements which the rule fails on,
+                         // if result == axs.constants.AuditResult.FAIL
+
+              /** @type {axs.AuditRule} */
+              rule  // The rule which this result is for.
+            }
+        """
+        if self.browser.name is not 'phantomjs':
+            raise NotImplementedError('Only supported in PhantomJS')
```
benpatterson: nit: just a bit more detail in the error message. Like "accessibility audit only supported..."

bok_choy/web_app_test.py
benpatterson: Ohhhhh I see it further down. Sorry. Ignore.

tests/test_accessibility.py
```diff
@@ -0,0 +1,26 @@
...
+Test accessibility auditing.
+
+You can run under phantomjs like this:
+SELENIUM_BROWSER=phantomjs ./run_tests.sh test_accessibility.py
+"""
+from bok_choy.web_app_test import WebAppTest
+from .pages import VisiblePage
+
+
+class AccessibilityTest(WebAppTest):
```
benpatterson: Maybe also add a test for the browser exception (e.g., you're not using phantomjs). I figure that can be mocked for travis.

benpatterson:
Overall I think it looks good. Couple questions:
* Will sphinx pick up the doc for us? I just want to be sure users have a 'how-to' for accessibility testing
* Should we make a global setting to always audit given pages? Er, assuming it is cheap. 
* How will we report results when we're running multiple tests? Does the ghostdriver log summarize at the end of the test session somehow?

tests/test_accessibility.py
jzoldak: Travis runs the tests only under firefox. Thus implicitly the test_axs_audit testcase will assert on the exception. And the successful audit is not actually performed under travis, which is why I put in that instruction on how to run under phantom locally.

You're right that it would be better if all this was explicit. And if travis runs the test under phantom.

I tried running all the tests under phantom and some fail. I feel that fixing them is outside the scope of this PR. I'll look into adding another commit which will run selected (maybe just this one) tests under phantom as well as the entire suite under firefox.

jzoldak:
* Yes sphinx will pick up the docs, but only on official releases - which either @wedaly or I historically have done. I plan to make a new release after we get this change in, as I will want to bump up the version that edx-platform is using.

* My vision is not to audit automatically, as the interpretation of accessibility standards can be a bit subjective, and also unfortunately pretty much all pages in the real world have at least one violation on them. :( Instead I was envisioning that acceptance testcases are written using the framework to get the report and then make assertions on that report, as exemplified in the test_axs_audit testcase.  With a little bit of usage, I'm anticipating that we will discover the usefulness of adding in some assertion helper methods, but I'm not sure at this point what they should be.

* No results are logged, instead we will use assertions. See above bullet point.

clytwynec:
I'm still a bit confused about how this will play out with assertions.   If we can't audit automatically because the standards are subjective, it seems strange to use assertions rather than storing a report so that it can be reviewed.

jzoldak:
@benpatterson -
Having done more exploration on the automatic auditing / logging of results, I can't wrap my head around how this option would work in practice. This doesn't fit into the paradigm of automated tests, where each test discretely needs to either pass or to fail; you don't look at the printout of every testcase. And gathering up and printing out results over a series of pages doesn't make sense, as the results won't be tied to any objective description out of context, like a page URL because they can be dependent on interactions that you've had with the page. E.g. a certain test may have clicked an element that caused other elements to be hidden, classes applied, etc. before running the report.

@clytwynec -
I anticipated that the assertions of each testcase would be written knowing the current audit results at that time, and to either intentionally fail or to catch future regressions. For example, I have a page on which I know that three specific elements fail the rule AX_ARIA_04 (badAriaAttributeValue). I would write a testcase asserting that these three still fail, and no other elements fail, and that no other rules are violated.

benpatterson:
@jzoldak Maybe we think about it more like linting than we do testing? That is, we have a target level of compliance (could be 100%) that is cross-tabulated with specific rules. And we have a count of violations?

jzoldak:
Incorporated on- and off-line feedback, now includes both an active and passive auditing ability. And documentation. :)
@clytwynec @benpatterson this PR is ready for re-review.

docs/accessibility.rst
```diff
@@ -0,0 +1,182 @@
...
+
+The general methodology for enabling accessibility auditing is:
+
+* Define the accessibility rules to check for a page
+* (optional) Define the scope for accessibility auditing of a page
+* Trigger the audits either:
+
+  * Actively, by calling the Page Object class's do_axs_audit method and then
+  asserting on the results returned.
+  * Passively, by setting the VERIFY_ACCESSIBILITY environment variable to True.
```
clytwynec: This bullet isn't getting formatted properly.  White space issue maybe?

docs/accessibility.rst
```diff
@@ -0,0 +1,182 @@
...
+  This will trigger an accessibility audit whenever a Page Object's wait_for_page
+  method is called, and raise an AccessibilityError if errors are found on the page.
+
+*Note that accessibility auditing is only supported with PhantomJS as the browser.*
+
+Define the accessibility rules to check for a page
+--------------------------------------------------
+
+A page object's list of accessibility rules to audit for that page are defined
+in the axs_audit_rules method. The code for the available rules can be found
```
clytwynec: I was confused about what  'code' means here.   The javascript code has code values for each rule in it. (e.g. BadAriaAttributeValue.js has `code: 'AX_ARIA_04'` in the addRule call) 

I was then confused why ''badAriaAttributeValue'' was being used in your example below.    It might be helpful to specify "javascript code" here, and to mention that the important part to look for is the 'name' attribute used in addRule.

docs/accessibility.rst
```diff
@@ -0,0 +1,182 @@
...
+    def axs_audit_rules(self):
+        return []
+
+To check a single rule, or multiple rules, configure the page object's
+axs_audit_rules method to return a list with the name(s) of the rule(s).
+
+.. code-block:: python
+
+    def axs_audit_rules(self):
+        return ['badAriaAttributeValue', 'imagesWithoutAltText']
```
clytwynec: Is there a way to say "use all of the rules _except_ for this subset'' ?

docs/accessibility.rst
```diff
@@ -0,0 +1,182 @@
+Accessibility Testing
+======================
+
+The bok-choy framework includes the ability to test web pages for accessibility
+using `Google Accessibility Developer Tools`_.
+
+The audit rules that are checked and optionally the scope of the audit
+within the webpage itself are defined on each Page Object's definition.
+
+The general methodology for enabling accessibility auditing is:
```
clytwynec: I just noticed that these bullet points actually outline the sections of the rest of the page.  Can these be linked to the corresponding sections?

run_tests.sh
```diff
@@ -39,7 +39,7 @@ python -m tests.http_server $SERVER_PORT 2> /dev/null &
 echo "Running test server on port $SERVER_PORT"
 
 echo_task "Run unit test suite"
-coverage run -m nose $REPO_ROOT/tests/$1
+coverage run -m nose $REPO_ROOT/tests/$1 --nocapture
```
clytwynec: Why is this change needed?

jzoldak: oh snaps, didn't mean to check that in! (It causes print statements to not be swallowed up by nose, so I use it for debugging.)

bok_choy/page_object.py
```diff
@@ -2,21 +2,28 @@
...
+
+import requests
 from selenium.common.exceptions import WebDriverException
 
 from .query import BrowserQuery
 from .promise import Promise, EmptyPromise, BrokenPromise
 
 
+AXS_FILE = 'vendor/axs_testing.js'
+AuditResults = namedtuple('AuditResults', 'errors, warnings')
```
clytwynec: Is there a reason to use `'errors, warnings'` over `['errors', 'warnings']` here?

docs/accessibility.rst
jzoldak: The js that we're using does have a way to do this, but I only plumbed in the rules to run, not the rules to ignore. Was planning on leaving that for a future implementation. See [Audit.js](https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/src/js/Audit.js#L48-60)

That said, I'm realizing it'd probably be better to rename this "audit_rules_to_run" right now in this first implementation, to leave space for adding "audit_rules_to_ignore" later.

docs/accessibility.rst
jzoldak: omg I can't believe you're making me learn rst.

clytwynec: I _think_ you can just use the same syntax as the other links, but use the heading texts. (  \`My header text\`_  )

bok_choy/page_object.py
```diff
@@ -173,6 +187,7 @@ def __init__(self, browser):
             PageObject
         """
         self.browser = browser
+        self.verify_accessibility = bool(os.environ.get('VERIFY_ACCESSIBILITY'))
```
clytwynec: What happens if I `export VERIFY_ACCESSIBILITY=False` or `export VERIFY_ACCESSIBILITY=0`?   I think that would result in self.verify_accessibility being True.

bok_choy/page_object.py
```diff
@@ -500,3 +520,130 @@ def wait_for_element_invisibility(self, element_selector, description, timeout=6
 
         """
         self.wait_for(lambda: not self.q(css=element_selector).visible, description=description, timeout=timeout)
+
+    def axs_audit_rules(self):
```
benpatterson: Not a must, but would it make sense to have these in a different class that gets inherited by PageObject?

bok_choy/page_object.py
```diff
@@ -500,3 +520,130 @@ def wait_for_element_invisibility(self, element_selector, description, timeout=6
...
+        See https://github.com/GoogleChrome/accessibility-developer-tools/tree/master/src/audits
+
+        E.g. return ['badAriaAttributeValue']
+        An empty list means to check for all available rules.
+        None means that no audit should be done for this page.
+        """
+        return None
+
+    def axs_scope(self):
+        """
```
benpatterson: This is awesome; so glad you built this in here.

bok_choy/page_object.py
```diff
@@ -500,3 +520,130 @@ def wait_for_element_invisibility(self, element_selector, description, timeout=6
...
+            NotImplementedError if you are not using PhantomJS
+            RuntimeError if there was a problem getting the report
+
+        Returns:
+            A list (one for each browser session) of namedtuples with 'errors' and 'warnings'
+            fields whose values are the errors and warnings returned from the audit.
+
+            None if the page object has no rules defined to check.
+        """
+        if self.browser.name != 'phantomjs':
```
benpatterson: Should we raise an exception when someone has asked for verify accessibility, but has no rules?

bok_choy/page_object.py
```diff
@@ -500,3 +520,130 @@ def wait_for_element_invisibility(self, element_selector, description, timeout=6
...
+            msg = 'Accessibility auditing is only supported with PhantomJS as the browser.'
+            raise NotImplementedError(msg)
+
+        rules = self.axs_audit_rules()
+        if rules is None:
```
benpatterson: Sorry this is where I should have asked the question...should we raise an exception when no rules are defined, but someone's asking for an audit?

docs/accessibility.rst
```diff
@@ -0,0 +1,182 @@
...
+
+This will trigger an audit, using the rules (and optionally the scope) set in
+the page object definition, whenever a call to wait_for_page() is made.
+
+In the case of the test_button_click_output testcase in the above example,
+an audit will be done at the end of the visit() and click_button() method calls,
+as each of those will call out to wait_for_page().
+
+If any assessibility errors are found, then the testcase will fail with an
+AccessibilityError.
```
benpatterson: Could you give an example here? Just to make it easy-peasy for python novices to understand.

vendor/axs_testing.js
```diff
@@ -0,0 +1,1710 @@
+/*
```
benpatterson: Not a must: should the path be vendor/google/axs_testing.js?

tests/test_accessibility.py
```diff
@@ -0,0 +1,128 @@
...
+        self.assertEqual(2, len(result.errors))
+        self.assertEqual(result.errors[0], self.ax_aria_04_errors)
+        self.assertEqual(result.errors[1], self.ax_aria_01_errors)
+
+        self.assertEqual(0, len(result.warnings))
+
+    @patch('tests.pages.AccessibilityPage.axs_scope')
+    @patch('tests.pages.AccessibilityPage.axs_audit_rules')
+    def test_axs_audit_limit_scope(self, mock_rules, mock_scope):
+        # Limit the scope to the div with AX_ARIA_04 violations
```
benpatterson: Nit...make these docstrings?

benpatterson:
@jzoldak my big question is whether or not we need to raise an exception if someone asked for implicit testing but has set no rules. Or, perhaps that will come with the default (all) rules?

bok_choy/page_object.py
jzoldak: Oh, both formats mean the same time. I just picked one. Unless that's not what you meant?
https://docs.python.org/2/library/collections.html#collections.namedtuple
> The field_names are a sequence of strings such as ['x', 'y']. Alternatively, field_names can be a single string with each fieldname separated by whitespace and/or commas, for example 'x y' or 'x, y'.

clytwynec: The list format just _seems_ more pythonic to me, but I think it's just a personal preference.

bok_choy/page_object.py
clytwynec: I agree.  Would this allow us to do something like measure the accessibility of a specific xblock in the platform?

docs/accessibility.rst
jzoldak: ok done.

bok_choy/page_object.py
jzoldak: Yes. Or test the content of courseware, ignoring the surrounding lms chrome (which will be tested in the context of the platform itself), etc.

jzoldak:
@benpatterson after talking this through with @clytwynec, changed the implementation to check all rules by default. This will cover the use case "As a developer I turn on implicit rules checking and see tons of of errors as my tests traverse my application."

We think it will be useful, and can later implement, overriding that default via environment variable.

And also I added a warning if rules for a page are checked, either actively or implicitly, and the page object is configured to not check any rules.

benpatterson:
@jzoldak just to clarify, that would be "check all rules by default [when I set the env variable]", is that right?

In other words..this sounds great. Assuming that I could still--by contrast--have my own set of rules to use on the tests I specify...

jzoldak:
@benpatterson The hierarchy we are currently proposing:
* (TBD, not implemented, still under discussion) takes precedence: appropriately named feature flag saying no matter what I have defined elsewhere override and run with these specified rules
* (currently implemented): defined on page object
* (TBD, not implemented): appropriately named feature flag saying if not specified on the page object, use these rules (overriding the default below)
* (currently implemented): default to using all rules

benpatterson:
Great summary. SGTM.

benpatterson:
:+1:

cptvitamin:
very excited to see progress on this @jzoldak @clytwynec @benpatterson.  Let me know when you have something for me to review.  Looking forward to seeing this in action.

clytwynec:
:+1:

PR #90, opened 2015-05-12
Call super in setUp.

cahrens:
@cpennington please review

Note that I haven't tried pulling this commit into edx-platform yet. I don't know if there is a risk of existing tests breaking.


cpennington:
:+1:

PR #91, opened 2015-05-13
Version 0.3.4

jzoldak:


jzoldak:
@clintonb 
FYI If you want to change your upstream reference that needed your basic auth fix - I created a new release, published to pypi.

clintonb:
Thanks!

PR #92, opened 2015-05-13
Enable linting on this repo and fix all pylint errors

jzoldak:
@clytwynec


bok_choy/browser.py
```diff
@@ -206,8 +206,7 @@ def _local_browser_class(browser_name):
     """
 
     # Log name of local browser
-    LOGGER.info(
-        "Using local browser: {0} [Default is firefox]".format(browser_name))
+    LOGGER.info("Using local browser: %s [Default is firefox]", browser_name)
```
clytwynec: In the platform, I thought we favored using `.format()`.   Will bok-choy not follow that?

clytwynec:
One question, but otherwise :+1:

bok_choy/browser.py
jzoldak: This does look like string formatting, but it's actually not. Confusing for sure.

The logger method call takes a format string, and optionally args and kwargs. See https://docs.python.org/2/library/logging.html#logging.Logger.debug

We used to be passing in only a fully-formed string as the msg. There's a newish pylint directive that tries to enforce that it's more optimal to pass in an actual format string and the arguments rather than using string formatting to compose the string either old school (%) or new school (.format) style ahead of time, because sometimes that calculation will be wasted. For example if you are passing stuff to Logger.info but you've configured the logger to only take action on CRITICAL level then the system need never format the string.

If that makes any sense.

clytwynec: Ok, I see. That makes sense.

PR #95, opened 2015-05-21
Add axs_audit_rules_to_ignore

clytwynec:


clytwynec:
@jzoldak @benpatterson Please review

tests/test_accessibility.py
```diff
@@ -108,6 +112,196 @@ def test_axs_audit_limit_rules(self, mock_rules):
 
         self.assertEqual(0, len(result.warnings))
 
+    @patch('tests.pages.AccessibilityPage.axs_audit_rules_to_run')
+    def test_axs_audit_run_multiple_rules(self, mock_rules):
+        # Tests that other all errors are reported if checking more than one.
```
benpatterson: nit: typo...I think you want "...that all other errors..." instead of "...that other all errors..."

tests/test_accessibility.py
```diff
@@ -108,6 +112,196 @@ def test_axs_audit_limit_rules(self, mock_rules):
...
+        result = report[0]
+
+        self.assertEqual(1, len(result.errors))
+        self.assertEqual(result.errors[0], self.ax_aria_04_errors)
+
+        self.assertEqual(0, len(result.warnings))
+
+    @patch('tests.pages.AccessibilityPage.axs_audit_rules_to_ignore')
+    def test_axs_audit_ignore_multiple_rules(self, mock_ignore):
+        # Ignore multiple rules. In this test, we are explicitly ignoring all.
```
benpatterson: nit: I'd say "explicitly ignoring each" instead of "explicitly ignoring all." Just b/c when I first read it, I incorrectly think we have some "ignore_all_rules" capability...

tests/test_accessibility.py
```diff
@@ -108,6 +112,196 @@ def test_axs_audit_limit_rules(self, mock_rules):
...
+        # There was one page in this session
+        self.assertEqual(1, len(report))
+        result = report[0]
+
+        self.assertEqual(0, len(result.errors))
+        self.assertEqual(0, len(result.warnings))
+
+    @patch('tests.pages.AccessibilityPage.axs_audit_rules_to_ignore')
+    @patch('tests.pages.AccessibilityPage.axs_audit_rules_to_run')
+    def test_axs_audit_ignore_1_and_run_1(self, mock_rules, mock_ignore):
```
benpatterson: excellent. thanks for adding this test!

benpatterson:
Some nits but :+1:

tests/test_accessibility.py
```diff
@@ -29,11 +29,15 @@ class AccessibilityTest(WebAppTest):
     @patch.dict(os.environ, {'SELENIUM_BROWSER': 'phantomjs'})
     def setUp(self):
         super(AccessibilityTest, self).setUp()
+
+        # Name is badAriaAttributeValue
```
jzoldak: oh sorry you had to trace this back! maybe for the next person, in the comment be explicit: "Name" -> "Audit rule name"

tests/test_accessibility.py
```diff
@@ -108,6 +112,196 @@ def test_axs_audit_limit_rules(self, mock_rules):
 
         self.assertEqual(0, len(result.warnings))
 
+    @patch('tests.pages.AccessibilityPage.axs_audit_rules_to_run')
+    def test_axs_audit_run_multiple_rules(self, mock_rules):
+        # Tests that all errors are reported if checking more than one.
```
jzoldak: This seems more like a docstring than a comment.

tests/test_accessibility.py
```diff
@@ -108,6 +112,196 @@ def test_axs_audit_limit_rules(self, mock_rules):
...
+
+        mock_ignore.assert_called_once_with()
+        mock_rules.assert_called_once_with()
+
+        # There was one page in this session
+        self.assertEqual(1, len(report))
+        result = report[0]
+
+        # It is expected that the badAriaRole error would be ignored because it
+        # is in and axs_audit_rules_to_ignore, but the rule
```
jzoldak: typo - "and"

tests/test_accessibility.py
```diff
@@ -108,6 +112,196 @@ def test_axs_audit_limit_rules(self, mock_rules):
...
+        self.assertEqual(1, len(report))
+        result = report[0]
+
+        self.assertEqual(0, len(result.errors))
+        self.assertEqual(0, len(result.warnings))
+
+    @patch('tests.pages.AccessibilityPage.axs_audit_rules_to_ignore')
+    @patch('tests.pages.AccessibilityPage.axs_audit_rules_to_run')
+    def test_axs_audit_ignore_1_and_run_1(self, mock_rules, mock_ignore):
+        # Test that axs_audit_rules_to_ignore takes precedence over
```
jzoldak: This is not really testing precedence, but instead testing that you can specify both.

jzoldak:
Code LGTM. Only suggestions I have concern comments/docstrings. :+1:

clytwynec:
@jzoldak @benpatterson  I've updated for your comments.  Good to merge?

benpatterson:
:+1:

bok_choy/page_object.py
```diff
@@ -534,6 +534,18 @@ def axs_audit_rules_to_run(self):
...
+        """
+        List of rules to ignore for accessibility errors on the page.
+        See https://github.com/GoogleChrome/accessibility-developer-tools/tree/master/src/audits
+
+        E.g. return ['badAriaAttributeValue']
+        An empty list means to run rules as defined by axs_audit_rules_to_run.
+        Otherwise, if rules are listed here, they will be ignored even if
+        they are specified in axs_audit_rules_to_run.
+        """
+        return None
```
jzoldak: Maybe the default return value should be [] here rather than None, to keep consistency with axs_audit_rules_to_run. 

It is potentially confusing is that "None" has different meanings. If I'm understanding correctly, the value of axs_audit_rules_to_ignore overrides whatever you put in axs_audit_rules_to_run, UNLESS you set axs_audit_rules_to_run to None, in which case it doesn't matter what you put axs_audit_rules_to_ignore because no rules are checked.

clytwynec: Sort of...  it's a bit more complicated I think.  I see it as `[]` having different meanings, actually.

I think its more like  `rules_actually_run = set(axs_audit_rules_to_run) - set(axs_audit_rules_to_ignore)`. In `axs_audit_rules_to_ignore`, both [] and None are considered empty sets.  But, in `axs_audit_rules_to_run`,  `[]` (which is normally considered empty) is considered "the set of all rules" and None is considered an empty set.

Would it be less confusing to change it to `[]` here and not document the option of setting it to `None` though?

jzoldak: Yeah, that seems best to me. Set it to [] here and take None out of here and accessibility.rst (and the tests, if it is in any).

clytwynec: Done.

jzoldak:
:+1:

PR #96, opened 2015-06-10
Extend bok-choy to include needle's snapshot assertion capabilities

jzoldak:
TE-927

@benpatterson @clytwynec @kashifch


bok_choy/web_app_test.py
```diff
@@ -34,6 +36,57 @@ def __init__(self, *args, **kwargs):
...
+            cls.capture = True
+        if os.environ.get('NEEDLE_SAVE_BASELINE'):
+            cls.save_baseline = True
+        if os.environ.get('NEEDLE_CLEANUP_ON_SUCCESS'):
+            cls.cleanup_on_success = True
+
+        # Instantiate the diff engine
+        klass = import_from_string(cls.engine_class)
+        cls.engine = klass()
+        super(NeedleTestCase, cls).setUpClass()  # pylint: disable=bad-super-call
```
clytwynec: I'm not following this line.... 

Why not `super(WebAppTest, cls).setUpClass()`?

bok_choy/web_app_test.py
```diff
@@ -34,6 +36,57 @@ def __init__(self, *args, **kwargs):
         self.har_mode = getattr(
             self, 'har_mode', os.environ.get('BOK_CHOY_HAR_MODE', None))
 
+    @classmethod
+    def setUpClass(cls):
+        """
+        Override NeedleTestCase's setUpClass method so that it does not
+        start up the browser once for each testcase class.
+        Instead we start up the browser once per TestCase instance,
```
benpatterson: Thanks for catching that

bok_choy/web_app_test.py
```diff
@@ -34,6 +36,57 @@ def __init__(self, *args, **kwargs):
...
+        in the setUp method.
+        """
+        if os.environ.get('NEEDLE_CAPTURE'):
+            cls.capture = True
+        if os.environ.get('NEEDLE_SAVE_BASELINE'):
+            cls.save_baseline = True
```
benpatterson: Is this something we want to expose? I am very distrustful of "whatever the test does, just say it passes and record it for the next time."

bok_choy/web_app_test.py
```diff
@@ -34,6 +36,57 @@ def __init__(self, *args, **kwargs):
...
+    def get_web_driver(self):
+        """
+        Override NeedleTestCases's get_web_driver class method to return the WebDriver instance
+        that is already being used, instead of starting up a new one.
+        """
+        return self.browser
+
+    def set_viewport_size(self, width, height):
+        """
+        Override NeedleTestCases's set_viewport_size class method because we need it to operate
```
benpatterson: Could you add a bit of doc on what the intention of this is anyway? Not sure what viewport size means.

bok_choy/web_app_test.py
```diff
@@ -56,6 +109,11 @@ def setUp(self):
         tags = [self.id()]
         self.browser = browser(tags, self.proxy)
 
+        # Needle uses these attributes for taking the screenshots
+        self.driver = self.get_web_driver()
+        self.driver.set_window_position(0, 0)
```
benpatterson: How will this work if the thing being screenshot is far below in the page (i.e. out of view because we've scrolled up)?

tests/test_webapptest.py
```diff
@@ -0,0 +1,22 @@
...
+import os
+from bok_choy.web_app_test import WebAppTest
+from .pages import ImagePage
+
+
+class ScreenshotAssertTest(WebAppTest):
+    """
+    Test the integration with needle and its screenshot assertion capability.
+    """
+    def test_needle_screenshot(self):
```
benpatterson: Is there any way to test that assertScreenshot fails when it should fail? I'd like that negative test if possible.

bok_choy/web_app_test.py
jzoldak: That would run NeedleTestCase's setUpClass method as-is, which would start up the browser. This is the lines that we need from NeedleTestCase's setUpClass, including the call to it's super.

tests/test_webapptest.py
jzoldak: Oh yeah, good catch. I'll add that.

bok_choy/web_app_test.py
jzoldak: This is the mechanism that needle uses to create the screenshots. I'm writing some doc and will include a warning.

jzoldak: Ah, on further consideration I get what you're saying and you're right. I'll remove these.

clytwynec: Is the the mechanism that you would use to get the screenshot to compare?? Would we have to manually take and save the screenshots otherwise?

jzoldak: The way to capture the baselines is by running once with a nose attribute (--with-save-baseline). I just added a commit with doc on how you would do that.

bok_choy/web_app_test.py
jzoldak: Added to the documentation a reference to needle's doc which explains this.

bok_choy/web_app_test.py
jzoldak: This is copy/pasted out of Needle's classSetUp method. This sets the window position to 0,0 right after starting the browser. I'm not really sure why the original author included it but also I don't think it could hurt.

I'm not sure the answer to your question though, other than "however Needle handles it."

That said, I've thought about this a bit more and as a hardening enhancement I will code up a commit that adds a promise that the element exists and is visible before taking the screenshot.

jzoldak: Doh, that won't work. Screenshot assertions are a method of the WebAppTest object, and promises are related to PageObjects which WebAppTest objects don't know about.

Instead I'll update the example in the documentation to establish a pattern whereby you first assert that the element you want to capture the image of is visible, then do the screenshot assertion.

docs/image_assertions.rst
```diff
@@ -0,0 +1,117 @@
...
+            EdxHomePage(self.browser).visit()
+
+Add the call to assertScreenshot
+--------------------------------
+assertScreenshot() take two arguments: a CSS selector for the element to capture and a filename for the image.
+Here is the same testcase, with an assertion added that will ensure that the site logo for the edx.org home
+page has not changed.
+
+* The first argument, "img.site-logo" is the css locator for the element that we want to capture and compare.
+* The second argument, "edx_logo_header" is the filename (.png will be appended automatically) that will be used
```
benpatterson: Does that mean you can only use .png files? Should we point that out? (I realize you have to take a screenshot with the tool, but....)

benpatterson:
:+1:

docs/image_assertions.rst
```diff
@@ -0,0 +1,117 @@
+Image Assertions
```
clytwynec: Is this a standard term?  I would have expected it to be something like "Screenshot Comparisons".

clytwynec:
:+1:

benpatterson:
:+1:

PR #100, opened 2015-06-20
Add invisible property.

benpatterson:
Currently stating `not self.q(css=foo).visible` will return true when
an element doesn't even exist on the page. We need a truer 'invisible'
property that tests that, indeed, the element is there, but we can't see it.

### DIFF
--- a/tests/site/visible.html
+++ b/tests/site/visible.html
@@ -6,5 +6,6 @@
 <body>
     <div class="batman" style="display:none">Batman</div>
     <div class="superman">Superman</div>
+    <div class="joker" style="visibility:hidden">Joker</div>
 </body>
 </html>
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -521,7 +521,7 @@ class PageObject(object):
             timeout (float): Maximum number of seconds to wait for the Promise to be satisfied before timing out
 
         """
-        self.wait_for(lambda: not self.q(css=element_selector).visible, description=description, timeout=timeout)
+        self.wait_for(lambda: self.q(css=element_selector).invisible, description=description, timeout=timeout)
 
     def axs_audit_rules_to_run(self):
         """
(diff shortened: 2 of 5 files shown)


benpatterson:
@jzoldak @clytwynec 

I discovered this while debugging a test today. Running not self.q().visible will return a false, even when the property doesn't exist. So I thought we needed a better way to implement "not visible" that implied the element was still there.

clytwynec:
Do you have an example of where this happens?

It appears that `visible` does check that there is at least one element that exists before confirming that they are all visible. Is this really a bug in the `visible` function maybe, or maybe in the usage?  `invisible` is a nice addition though.

benpatterson:
@clytwynec if you add the following line to the end of the test_visible.py file, it will fail:
`self.assertEqual(self.page.is_invisible('foobar'), not self.page.is_visible('foobar'))`

clytwynec:
Ok.  So is_visible does return False when the element doesn't exist, but it ambiguous whether it is returning False because it is "present but hidden" or because it is "not present and therefore not visible".   And adding is_invisible provides a way to make that distinction.  Is that what you were saying?

benpatterson:
@clytwynec yes, exactly!

clytwynec:
:+1: 

(I think I might have misread the first comments and skipped over a 'not' in there.  Probably made this more confusing than it should have been)

tests/site/visible.html
```diff
@@ -6,5 +6,6 @@
 <body>
     <div class="batman" style="display:none">Batman</div>
     <div class="superman">Superman</div>
+    <div class="joker" style="visibility:hidden">Joker</div>
```
jzoldak: :gift: :boom:

jzoldak:
:+1:

benpatterson:
Ah not shipping it just yet. The old-usage (where invisibility could mean doesn't exist at all) was actually being used in some tests. Working that out here. https://github.com/edx/edx-platform/pull/8629

benpatterson:
Adjustments are queued up for platform to absorb this change. Merging.

PR #102, opened 2015-06-22
Fix packaging of vendor JS for accessibility testing

jzoldak:
@clytwynec @benpatterson - This fixes the root cause of accessibility tests breaking when updating the reference in the edx-platform repo to point to the version up at pypi. (It was unrelated to the needle / image diff feature.)

Thx @nedbat for the packaging advice. :)


clytwynec:
The vendor files were being included before because... we were installing from github with `-e`?

jzoldak:
@clytwynec - Yes.

clytwynec:
OK wow! I'm glad it wasn't something related to the feature itself! 

:+1:

PR #106, opened 2015-07-07
fix wait for ajax "jQuery is not defined" errors

clytwynec:
@jzoldak @benpatterson please review.

### DIFF
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -399,9 +399,11 @@ class PageObject(object):
         yield
 
     @unguarded
-    def wait_for_ajax(self):
+    def wait_for_ajax(self, timeout=30):
         """
-        Wait for all ajax requests to finish.
+        Wait for jQuery to be loaded and for all ajax requests to finish. Note
+        that we have to wait for jQuery to load first because it is used to
+        check that ajax requests are complete.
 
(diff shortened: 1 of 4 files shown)


bok_choy/page_object.py
```diff
@@ -399,9 +399,9 @@ def handle_alert(self, confirm=True):
         yield
 
     @unguarded
-    def wait_for_ajax(self):
+    def wait_for_ajax(self, timeout=30):
```
clytwynec: Added timeout so that the tests wouldn't have to wait 30 seconds to pass

jzoldak: Since the default for everything else seems to be 60, we should use that as the default here too.

jzoldak:
:+1:

benpatterson:
:+1:

bok_choy/page_object.py
clytwynec: I set it to 30 because it is the [default timeout in the Promise class](https://github.com/edx/bok-choy/blob/master/bok_choy/promise.py#L37) (which is where the timeout for this function was coming from before).

jzoldak: Ah! OK then 30 is great, carry on!

PR #107, opened 2015-07-08
update fix wait_for_ajax

clytwynec:
@jzoldak @benpatterson  It turns out that the previous solution still had a gap where we could see the same issue appear.  Here's the updated fix.

### DIFF
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -405,6 +405,10 @@ class PageObject(object):
         that we have to wait for jQuery to load first because it is used to
         check that ajax requests are complete.
 
+        Important: If you have an ajax requests that results in a page reload,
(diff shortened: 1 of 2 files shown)


jzoldak:
:+1: after tests pass.

bok_choy/page_object.py
```diff
@@ -405,6 +405,10 @@ def wait_for_ajax(self, timeout=30):
         that we have to wait for jQuery to load first because it is used to
         check that ajax requests are complete.
 
+        Important: If you have an ajax requests that results in a page reload,
```
benpatterson: nit: requests/request

benpatterson:
I had a nit but feel free to merge once tests pass. :+1:

PR #109, opened 2015-09-16
Make axe-core ruleset default in accessibility audit

clytwynec:
Refactors the accessibility auditing feature to allow use of either the google accessibility dev tools ruleset or dequelab's axe-core ruleset.  See updated docs/docstrings for details about implementation and usage.

The output for the google accessibility tools ruleset remains the same.  The output for the newly implemented axe-core ruleset looks something like the following:

```
AccessibilityError: URL 'http://localhost:8003/accessibility.html' has 6 errors:

Rule ID: aria-roles
Help URL: https://dequeuniversity.com/rules/axe/1.1/aria-roles

	Message: 
	Html: <div id="AX_ARIA_01_not_a_role" role="datepicker"></div>
	Target: [u'#AX_ARIA_01_not_a_role']

	Message: 
	Html: <div id="AX_ARIA_01_empty_role" role=""></div>
	Target: [u'#AX_ARIA_01_empty_role']

Rule ID: aria-valid-attr-value
Help URL: https://dequeuniversity.com/rules/axe/1.1/aria-valid-attr-value

	Message: Invalid ARIA attribute value: aria-hidden="yes"
	Html: <span id="AX_ARIA_04_bad" aria-hidden="yes">foo</span>
	Target: [u'#AX_ARIA_04_bad']

Rule ID: button-name
Help URL: https://dequeuniversity.com/rules/axe/1.1/button-name

	Message: Element does not have inner text that is visible to screen readers
	Html: <div id="AX_ARIA_01_01" role="button"></div>
	Target: [u'#AX_ARIA_01_01']

Rule ID: html-lang
Help URL: https://dequeuniversity.com/rules/axe/1.1/html-lang

	Message: The <html> element does not have a lang attribute
	Html: <html>
	Target: [u'html']

Rule ID: link-name
Help URL: https://dequeuniversity.com/rules/axe/1.1/link-name

	Message: Element does not have text that is visible to screen readers
	Html: <a href="#" style="color: white; background-color: white;"></a>
	Target: [u'#bad-link > a']
```


clytwynec:
@benpatterson @jzoldak  Please review when you get a chance.  I still have to update accessibility.rst and figure out why some of these tests are failing (they are failing on master now too), but it's a big PR, so I wanted to give you a heads up and let you get started reviewing if you have time.

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,161 @@
...
+    Allows auditing of a page for accessibility issues.
+
+    The ruleset to use can be specified by the environment variable
+    `BOKCHOY_A11Y_RULESET`. Currently, there are two ruleset implemented:
+
+    `axe_core`:
+
+        * Ruleset class: AxeCoreAudit
+        * Ruleset config: AxeCoreAuditConfig
+        * This is default the default ruleset.
```
jzoldak: typo - extra "default"

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,161 @@
...
+        ghostdriver_url = self.browser.service.service_url
+
+        # Get the session_id from ghostdriver so that we can inject JS into
+        # the page.
+        resp = requests.get('{}/sessions'.format(ghostdriver_url))
+        sessions = resp.json()
+
+        return ghostdriver_url, sessions
+
+    def _verify_file(self):
```
jzoldak: might want to rename this to _verify_file_exists

bok_choy/a11y/axe_core_ruleset.py
```diff
@@ -0,0 +1,296 @@
...
+from textwrap import dedent
+
+from .helpers import AccessibilityError
+from ..promise import Promise
+
+
+CUR_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+class AxeCoreAuditConfig(object):
```
jzoldak: A cleaner and more extensible implementation would be to think of axe-core and axs as plug-ins and derive them both from an [abstract base class](https://pymotw.com/2/abc/) with abstract methods that are overridden for check_rules, report_errors, etc.

bok_choy/a11y/axe_core_ruleset.py
```diff
@@ -0,0 +1,296 @@
...
+    def set_rules_file(self, path=None):
+        """
+        Sets `rules_file` to the passed file. Else it sets it to the
+        file included in vendor/axe-core/axe.min.js.
+
+        Args:
+            A filepath where the JavaScript for the ruleset can be found.
+
+        This is intended to be used in the case of using an extended or
+        modified version of the axe-core ruleset. The interface and repsonse
```
jzoldak: type response

bok_choy/a11y/axe_core_ruleset.py
```diff
@@ -0,0 +1,296 @@
...
+
+            exclude: a list of css selectors for elements that should not
+            be included in the audit.
+            include (optional): a list of css selectors for elements that
+            should be included in the audit. By, default, the entire document
+            is included.
+
+        Examples:
+
+            To indlude all items in `#main-content` except `#some-special-elm`::
```
jzoldak: typo include

bok_choy/a11y/axe_core_ruleset.py
```diff
@@ -0,0 +1,296 @@
...
+                (False, None) if the results aren't available.
+            """
+            resp = requests.post(
+                script_url,
+                data=json.dumps({"script": audit_results_script, "args": []})
+            )
+
+            try:
+                results = json.loads(resp.json().get('value'))
+            except Exception:  #pylint: disable=broad-except
```
jzoldak: what might cause an exception, and what exception type would be thrown?

bok_choy/a11y/axs_ruleset.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+AuditResults = namedtuple('AuditResults', 'errors, warnings')
+CUR_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+class AxsAuditConfig(object):
+    """
+    The `AxsAuditConfig` object defines the options available when
+    running and `AxsAudit`.    """
```
jzoldak: type "an" not "and"

bok_choy/a11y/axe_core_ruleset.py
clytwynec: Could you describe more about what exactly you have in mind?

jzoldak: An example in this codebase is the PageObject metaclass that page objects inherit from. Defined page objects override the abstract methods like is_browser_on_page, url, etc. 
Here you would make metaclasses for A11yAuditConfig and A11yAudit, with abstract methods that return NotImplementedError. That gives placeholders for the "plug-ins", in your case the Axs and AxeCore to override. And when a third ruleset gets added in it will follow the same pattern.
Hope this makes sense. We can discuss in person if it's more expedient.

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,228 @@
...
+    pass
+
+
+class A11yAuditConfig(object):
+    """
+    The `A11yAuditConfig` object defines the options available in an
+    accessibility ruleset.
+    """
+    def __init__(self, *args, **kwargs):
+        """
```
benpatterson: nit: Is pylint requiring a docstring? I think if it'll be empty, don't have one...

benpatterson:
I'm still groking this, but I'm confused between the `old` ruleset and the `new` ruleset. One of them has rules_to_ignore and the other one has just rules with an attribute of ignore. Similarly there's scope() and set_context() which I think are about the same...  

I think I saw in my stream some discussion of a base class with abstract methods? I think that's appropriate here, much like the PageObject has methods that are expected in derived classes. It would be good if the two rulesets were implemented with the same structure, and since we're now dealing with multiple rulesets, it's the right time to use the right base class structure. (again, still grok-ing, but I see how one of the ignore_rule methods or attributes might be an array, and the other is just one string? perhaps the base class expects an array and the derived class just pulls the one string out of it in one of those cases...but my point is the general concepts we're applying with how the rulesets are leveraged so they should look the same.) This will also make it easier to extend if we run into another ruleset, or end up with some customized ruleset.

Documentation looks pretty solid. I didn't read it for syntax or anything with this review, but it looked pretty good overall with examples, etc. I also saw you put in tests on the new ruleset which is great.

benpatterson:
Oh duh there already is a base class. Sorry...that's obvious to me now. But standardizing the methods with the abstract classes is the right call.

clytwynec:
@benpatterson @jzoldak  For the config classes, I made rules_file standard, but not the other methods.  I'm very hesitant to standardize the other methods because they are ruleset specific.  For example, while the google axs ruleset allows you to set both rules_to_ignore and rules_to_run, you can't use both in the axe-core ruleset.  I tried to stay as close the the ruleset's javascript implementation of the options/config rather than forcing it to fit some other model. Doing that makes the ruleset's documentation (the stuff on the google-dev-tools repo or the deque repo) more relevant.

benpatterson:
@clytwynec I understand your point. However, I'm looking at this from the perspective of a non-accessibility developer...I would not know the finer nuances between the rulesets. The point of the abstraction is to prevent me from needing to know those nuances.

For the example you describe, can't you just enforce that a person does not pass in both ignore rules and set rules? Like, if someone does not pass in True or False to the ignore attribute, then the user encounters an error? How about scope vs context? I didn't see a big difference between those.... Or alternatively, have the set_rules attribute accept a JSON doc or some other thing that is flexible enough to be used by either one?

I understand your point but I'd like to figure out something that makes this easier to maintain, and for an accessibility newbie to adopt. As it stands I think we need more standardization and I really don't think we are that far off....

clytwynec:
@benpatterson @jzoldak  I've updated this PR to address your comments.  I still have to update accessibility.rst, but I'd like to keep that as a separate PR.

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,253 @@
...
+            A filepath where the JavaScript for the ruleset can be found.
+
+        This is intended to be used in the case of using an extended or
+        modified version of the ruleset. The interface and response
+        format are expected to be unmodified.
+        """
+        if path:
+            self.rules_file = os.path.abspath(path)
+
+    def set_rules(self, rules, ignore=False, include_type="rule"):
```
benpatterson: Do these need the abstract method decorator?

benpatterson:
I took a quick fast and it looks better now...will take a closer look later, hopefully today.

bok_choy/a11y/a11y_audit.py
clytwynec: Nice catch.  They don't need it necessarily, but it's nice.  I just updated this and a couple of other methods that should have it too.

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,267 @@
...
+        """
+        Overrides the default scope (part of the DOM) to inspect.
+
+        Raises:
+
+            `NotImplementedError` if this isn't overwritten in the ruleset
+                specific implementation.
+        """
+        raise NotImplementedError(
+            "The ability to specify has not been implemented."
```
benpatterson: nit: I think this needs a noun, such as "The ability to specify an area of the DOM..." or "the ability to specify where rules should be assessed..."

bok_choy/a11y/axe_core_ruleset.py
```diff
@@ -0,0 +1,321 @@
...
+                )
+
+            To run all rules except for "bad-link" and "color-contrast"::
+
+                page.a11y_audit.config.set_rules(
+                    ["bad-link", "color-contrast"],
+                    ignore=True
+                )
+
+            To run only WCAG 2.0 Level A rules::
```
benpatterson: Great example.

benpatterson:
This LGTM :+1:

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,267 @@
...
+            config: (optional) A11yAuditConfig or subclass of A11yAuditConfig
+        """
+        super(A11yAudit, self).__init__(*args, **kwargs)
+        self.url = url
+        self.browser = browser
+        self.config = config or self.default_config
+
+    def _phantomjs_setup(self):
+        """
+        Checks that phantomjs is being used.
```
jzoldak: Verifies that phantomjs is being used and returns the ghostdriver URL and session information.

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,267 @@
...
+        ghostdriver_url = self.browser.service.service_url
+
+        # Get the session_id from ghostdriver so that we can inject JS into
+        # the page.
+        resp = requests.get('{}/sessions'.format(ghostdriver_url))
+        sessions = resp.json()
+
+        return ghostdriver_url, sessions
+
+    def _verify_file_exists(self):
```
jzoldak: This should probably be named _verify_rules_file_exists, as it doesn't take a filename as a parameter and it is specifically checking self.rules_file's existence.

tests/test_accessibility.py
```diff
@@ -93,15 +127,20 @@ def test_axs_audit_limit_scope(self, mock_scope):
...
+        with self.assertRaises(NotImplementedError):
+            page.a11y_audit.config.set_scope(exclude='')
+
+    def test_axs_audit_limit_rules(self):
         page = AccessibilityPage(self.browser)
-        page.visit()
-        report = page.do_axs_audit()
 
-        mock_rules.assert_called_once_with()
+         # Limit the rules checked to AX_ARIA_01
```
jzoldak: This line has pep8 violations due to an extra space in indentation.

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,267 @@
...
+    def report_errors(audit, url):
+        """
+        Args:
+
+            audit: results of an accessibility audit.
+            url: the url of the page being audited.
+
+        Raises:
+
+            `AccessibilityError` should be raised when errors are found in the
```
jzoldak: I'd take out the words "should be raised"

bok_choy/a11y/axs_ruleset.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    def __init__(self, *args, **kwargs):
+        super(AxsAuditConfig, self).__init__(*args, **kwargs)
+        self.rules_file = None
+        self.scope, self.rules_to_run, self.rules_to_ignore = None, None, None
+        self.rules_file = os.path.join(
+            os.path.split(CUR_DIR)[0],
+            'vendor/google/axs_testing.js'
+        )
+        self.set_rules([], ignore=False)
```
jzoldak: ? Maybe this line isn't supposed to be here?

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Args:
+            browser: A phantomjs browser instance
+            url: URL of the page to test
+            config: (optional) A11yAuditConfig or subclass of A11yAuditConfig
+        """
+        super(A11yAudit, self).__init__(*args, **kwargs)
+        self.url = url
+        self.browser = browser
+        self.config = config or self.default_config
```
jzoldak: I must be missing something. Where would self.default_config be initialized prior to __init__?

bok_choy/a11y/axs_ruleset.py
clytwynec: The first one sets rules_to_run, second sets rules_to_ignore.

bok_choy/a11y/a11y_audit.py
clytwynec: @jzoldak `self.default_config` is a property of a `A11yAudit` (see [line 202](https://github.com/edx/bok-choy/pull/109/files#diff-8650032c466196a84557f7f5368250d5R222))

bok_choy/a11y/axs_ruleset.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def set_rules(self, rules, ignore=False, include_type="rule"):
+        """
+        Sets `rules_to_run` or `rules_to_ignore` to the passed list of rule IDs.
+
+        Args:
+
+            rules: List of rules to ignore or run for accessibility errors on
+                the page. If `ignore=False`, an empty list means to check for
+                all available rules and `None` means that no audit should be
```
jzoldak: I'm not sure what this means. If what is None, ignore=None?

bok_choy/a11y/axs_ruleset.py
```diff
@@ -0,0 +1,239 @@
...
+                done for this page. If `ignore=True`, an empty list means to
+                run rules all otherwise enabled rules and any rules listed here
```
jzoldak: I think there's an extra word "rules" in this sentence.

bok_choy/a11y/axs_ruleset.py
clytwynec: the value of  `rules`

clytwynec:
@jzoldak I just updated to address your comments.

bok_choy/a11y/axs_ruleset.py
jzoldak: Oh, I think I get it - if rules is None.

bok_choy/a11y/axs_ruleset.py
```diff
@@ -0,0 +1,239 @@
...
+            the errors and warnings returned from the audit.
+
+            None if config has rules_to_run set to None.
+
+        __Caution__: You probably don't really want to call this method
+        directly! It will be used by `A11yAudit.do_audit` if using this ruleset.
+        """
+        if config.rules_to_run is None:
+            msg = 'No accessibility rules were specified to check.'
+            print msg
```
jzoldak: Use the logger rather than print.

bok_choy/a11y/axs_ruleset.py
```diff
@@ -0,0 +1,239 @@
...
+    @staticmethod
+    def get_errors(audit):
+        """
+        Args:
+
+            audit: results of `AxsAudit.do_audit()`.
+
+        Returns:
+
+            AuditResults: a namedtuple (`errors`, `warnings`).
```
jzoldak: This returns a list or errors.

bok_choy/a11y/helpers.py
```diff
@@ -0,0 +1,10 @@
+"""
+Common helper methods for accessibility audit functionality.
+"""
+
+
+class AccessibilityError(Exception):
```
jzoldak: It looks like this file can be removed now as AccessibilityError is defined in a11y/a11y_audit.py.

bok_choy/a11y/axs_ruleset.py
```diff
@@ -0,0 +1,239 @@
...
+        else:
+            self.rules_to_run = rules
+
+    def set_scope(self, include=None, exclude=None):
+        """
+        Sets `scope`, the "start point" for the audit.
+
+        Args:
+
+            include: A selector specifying the element which contains the
```
jzoldak: We need to normalize this method across the 2 implementations. That is, AxsAuditConfig requires the set_scope method's include argument to be a string representing a css selector, and AxeCoreAuditConfig requires a list of strings. So if for example you have a testcase class that initializes the audit type in the setup method which has a bunch of testcases that set the include scope, it will not be straightforward to switch back and forth between the rulesets. E.g. you can't ddt to run them both.

clytwynec: Ah. Yea, I actually updated this method to take a list, and I updated the example but missed this part of the docstring. I'll fix it.

bok_choy/a11y/a11y_audit.py
jzoldak: Ah, OK I get it now thx.

bok_choy/a11y/axs_ruleset.py
jzoldak: Could you please change this docstring to read more clearly? The docstring for set_rules in AxeCoreAuditConfig is great, and really easy to understand.
Maybe for this one break the explanation out into what it means if rules is None, and what it means if rules is an empty list. And in each of those cases explain what it means if ignore=False vs. ignore=True.

bok_choy/a11y/a11y_audit.py
```diff
@@ -0,0 +1,267 @@
...
+
+        This is intended to be used in the case of using an extended or
+        modified version of the ruleset. The interface and response
+        format are expected to be unmodified.
+        """
+        if path:
+            self.rules_file = os.path.abspath(path)
+
+    @abstractmethod
+    def set_rules(self, rules, ignore=False, include_type="rule"):
```
jzoldak: The ignore argument being a boolean flag whose opposite means run was really tripping me up. 
I think the code would be more readable if instead there was a parameter named something like "action" whose argument values could either be "run" or "ignore".

clytwynec: I was thinking of it more as a "by default, `set_rules` means to run, and optionally you change it to ignore these rules instead.  I personally think that adding the boolean `ignore=True` will be easier to recall in practice than `action="ignore"`.

clytwynec: @jzoldak @benpatterson  Any further thoughts on this?

jzoldak:
:+1:

benpatterson:
:+1:

PR #110, opened 2015-09-18
Check the loading state of the document in wait_for_page

jzoldak:
@clytwynec @benpatterson 
This is an attempt to reduce the likelihood of flakiness due to synchronization/timing in tests.

### DIFF
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -342,6 +342,20 @@ class PageObject(object):
         Raises:
             BrokenPromise: The timeout is exceeded without the page loading successfully.
         """
+        def _is_document_ready():
+            """
+            Check the loading state of the document to ensure the document and all sub-resources
+            have finished loading (the document load event has been fired.)
+            """
+            return self.browser.execute_script(
+                "return document.readyState=='complete'")
+
+        EmptyPromise(
+            _is_document_ready,
+            "The document and all sub-resources have finished loading.",
+            timeout=timeout
+        ).fulfill()
+
         result = Promise(
             lambda: (self.is_browser_on_page(), self), "loaded page {!r}".format(self),
             timeout=timeout


jzoldak:
FWIW the tests in this repo all passed without changes, as did the 700+ tests in the edx-platform repo.

jzoldak:
also FWIW I ran the bok-choy tests in the edx-platform repo with this change, and also with a temporary change to make anything decorated with flaky to run 10 times and need to pass all 10. There was only 1 failure, which was this test: common.test.acceptance.tests.studio.test_studio_course_team.CourseTeamPageTest.test_admins_can_delete_other_users

[Here is the jenkins test report](https://build.testeng.edx.org/job/edx-platform-bok-choy-pr/4942/testReport/)

jzoldak:
@clytwynec @benpatterson thoughts?

bok_choy/page_object.py
```diff
@@ -338,6 +338,20 @@ def wait_for_page(self, timeout=30):
...
         """
+        def _is_document_ready():
+            """
+            Check the loading state of the document to ensure the document
+            and all sub-resources have finished loading.
+            """
+            return self.browser.execute_script(
+                "return document.readyState=='complete'")
+
+        EmptyPromise(
```
benpatterson: More of a nit, but do we want to use `wait_for` rather than `EmptyPromise`?

benpatterson:
My only concern is whether or not we'll get bitten by a 3rd-party resource that is not important for the tests, but keeps spinning as we await a response.

jzoldak:
@benpatterson Because it's wrapped in a promise, if a 3rd party resource prevents the document readyState from being complete then the wait_for_page will fail with a BrokenPromise after the timeout (default 30 seconds).

bok_choy/page_object.py
```diff
@@ -338,6 +338,20 @@ def wait_for_page(self, timeout=30):
...
+            _is_document_ready,
+            "The document load event has been fired.",
+            timeout=timeout
```
clytwynec: Not a must, but could this say something more similar to the docstring in _is_document_ready, like `the document and all sub-resources have finished loading`?

bok_choy/page_object.py
jzoldak: I could go either way, but I guess I used an EmptyPromise here to be similar to the Promise in the next statement below.

jzoldak:
@clytwynec @benpatterson OK to merge?

clytwynec:
This looks good to me.  :+1:

PR #112, opened 2015-09-28
refactor A11yAuditConfig.set_rules

clytwynec:
@jzoldak @benpatterson  I refactored the `set_rules` method as we were talking about this morning, partly just as an exercise for myself to see it in place.   Seeing the implementations side by side,  I think either way is fine.   It seems to me that both implementations are probably going to have confusing points for end users, but if this way is going to be easier to maintain down the road, then it is probably good to update.   Let me know what you think.

### DIFF
--- a/docs/accessibility.rst
+++ b/docs/accessibility.rst
@@ -30,35 +30,33 @@ This can be updated after instantiating the page object to be tested via the
 ``set_rules`` method.
 
 The default is to check all the rules. To set this explicitly, pass an empty
-list to ``set_rules``.
+dictionary to ``set_rules``.
 
 .. code-block:: python
 
-    page.a11y_audit.config.set_rules([])
+    page.a11y_audit.config.set_rules({})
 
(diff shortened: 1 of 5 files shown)


docs/accessibility.rst
```diff
@@ -30,35 +30,34 @@ This can be updated after instantiating the page object to be tested via the
...
 
-    page.a11y_audit.config.set_rules(
-        ['badAriaAttributeValue', 'imagesWithoutAltText']
-    )
+    page.a11y_audit.config.set_rules({
+        "apply": ['badAriaAttributeValue', 'imagesWithoutAltText'],
+    })
 
 To skip checking a specific set of rules on a particular page, pass the list
 of the names of the rules as the first argument and ``ignore=True`` as the
```
jzoldak: this line needs to be updated too.

jzoldak:
Looks great to me! Thanks. :+1:

benpatterson:
I also think this looks good. Thanks for giving it another pass. :+1:

PR #114, opened 2015-09-28
fix a11y api autodocs

clytwynec:
@benpatterson @jzoldak The new docs didn't update like I expected.  This fixes them.


benpatterson:
:+1: thanks for catching that.

jzoldak:
:+1:

PR #116, opened 2015-10-02
allow A11yAudits to run on browsers other than phantomjs

clytwynec:
I've run into some issues with phantomjs when running the edx-platform tests.  I think these could easily affect others that may want to use the a11y feature as well.

__The two main issues I uncovered:__
*  PhantomJS 1.9.8 runs on an older version of JavaScriptCore, which lacks the method `Function.prototype.bind` (a pretty commonly used method).  PhantomJS 2.0 has the newer version of JavaScriptCore, so this wouldn't be an issue if we could upgrade.  Unfortunately, there is still no binary package available for Linux for version 2.0.  This issue will likely continue to be a problem for a while. 
* PhantomJS doesn't support HTML5 video player (the favored player in platform) or Flash (the fallback player in platform).  It doesn't appear that this will be supported in the future.

__The solution here:__  An implementation of A11yAudit that doesn't _require_ PhantomJS.

__Implementation:__  Previously, we performed an A11yAudit by using PhantomJS' injectJs method to inject the auditing library's javascript code, and then sending a request to PhantomJS to execute the JS audit code and return the results.  The changes here bypasses this by using the `browser.execute_script` method, including the necessary library code in the script. 

@benpatterson @jzoldak  Please review.

FYI, the decrease in coverage is because I removed some large chunks of code.  The uncovered lines are the same ones as before --  the `raise NotImplementedError` lines.

### DIFF
--- a/tests/test_accessibility.py
+++ b/tests/test_accessibility.py
@@ -14,71 +14,32 @@ from bok_choy.promise import BrokenPromise
 from .pages import AccessibilityPage
 
 
-class NoAccessibilityTest(WebAppTest):
+class GoogleAxsAccessibilityTestMixin(object):
     """
-    Test unsupported accessibility audit configuration.
+    Test cases for axs ruleset accessibility audit integration.
     """
-    @patch.dict(os.environ, {'SELENIUM_BROWSER': 'firefox'})
-    def setUp(self):
-        super(NoAccessibilityTest, self).setUp()
+    # Audit rule name is badAriaAttributeValue
+    ax_aria_04_errors = u'{}{}{}{}'.format(
+        'Error: AX_ARIA_04 ',
+        '(ARIA state and property values must be valid) failed on the following ',
+        'element:\n',
+        '#AX_ARIA_04_bad')
 
-    def test_axs_audit_firefox(self):
-        page = AccessibilityPage(self.browser).visit()
-        with self.assertRaises(NotImplementedError):
-            page.a11y_audit.do_audit()
+    # Audit rule name is badAriaRole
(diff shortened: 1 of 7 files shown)


jzoldak:
@clytwynec 
Let's change the coverage config file in this repo to specifically not count lines that either `raise NotImplementedError` or are tagged with `pragma: no-cover`.

See [the example in the edx-platform](https://github.com/edx/edx-platform/blob/master/.coveragerc#L29).

jzoldak:
I think that docs/accessibility.rst also needs to be updated, as it says that PhantomJS is required.

tests/test_accessibility.py
```diff
@@ -14,71 +14,32 @@
 from .pages import AccessibilityPage
 
 
-class NoAccessibilityTest(WebAppTest):
+class AxsAccessibilityTestMixin(object):
```
jzoldak: nit: It would be easier to understand the test code and coverage if the mixin names were normalized. 
E.g. GoogleAxsTestMixin and AxeCoreTestMixin for BOKCHOY_A11Y_RULESET values google_axs and axe_core.

jzoldak:
@clytwynec @benpatterson I just changed the coverage settings on coveralls for this repo to require > 90% coverage, and a drop of no greater than 1% for a PR (it was set fail on any decrease for PRs).

Should be reflected in the next build.

clytwynec:
@jzoldak  I've updated with those changes.

jzoldak:
:+1:

benpatterson:
:+1:

PR #118, opened 2015-10-08
Add ability to include custom rules in an axe-core accessibility audit.

clytwynec:
This adds the ability to include custom rules or override existing implementations of a rule in an axe-core accessibility audit.  

The way it is implemented is that you can create a JS file that has the custom rules code and either:

* Set an env var (`BOKCHOY_A11Y_CUSTOM_RULES_FILE`) so that the rules in the file are included during all of the tests.
* Use `a11yAudit.config.customize_ruleset('path/to/custom_rules_file.js')` to include it in a specific test.

See the example of what a custom rules file looks like in `tests/a11y_custom_rules.js`.


clytwynec:
@benpatterson @jzoldak  Sorry for yet another PR.  This one isn't urgent but would be a nice addition.  Please review when you have time.

benpatterson:
@clytwynec makes sense. 
Some questions...
- Could you add a test(s) for when the file is not found?
- It wasn't obvious to me, but do any of the tests cover extending the rules (e.g., both the custom rules and the axe rules are exposing errors)?
- Do we need to refer anyone to documentation on how to write custom rules?

bok_choy/a11y/axe_core_ruleset.py
```diff
@@ -148,6 +151,53 @@ def set_scope(self, include=None, exclude=None):
...
+            To include the rules defined in `axe-core-custom-rules.js`::
+
+                page.a11y_audit.config.customize_ruleset(
+                    "axe-core-custom-rules.js"
+                )
+
+            Alternatively, use the environment variable `BOKCHOY_A11Y_CUSTOM_RULES_FILE`
+            to specify the path to the file containing the custom rules.
+
+        Documentation for how to write rules:
```
clytwynec: @benpatterson 
re: "Do we need to refer anyone to documentation on how to write custom rules?"

This is done here, along with a link to the example in the codebase.

tests/test_accessibility.py
```diff
@@ -312,6 +317,37 @@ def test_check_for_accessibility_errors(self):
         with self.assertRaises(AccessibilityError):
             self.page.a11y_audit.check_for_accessibility_errors()
 
+    def test_customize_ruleset_via_env_var(self):
+        with patch.dict(os.environ,
+            {'BOKCHOY_A11Y_CUSTOM_RULES_FILE': 'tests/a11y_custom_rules.js'}):
+            self.page.a11y_audit.config.customize_ruleset()
+            self._do_audit_and_check_errors(7)
+
+    def test_customize_ruleset_via_config_param(self):
```
clytwynec: @benpatterson 
re: "It wasn't obvious to me, but do any of the tests cover extending the rules (e.g., both the custom rules and the axe rules are exposing errors)?"

This test and the one right above have errors resulting from both the custom ruleset and the regular ruleset. (Without the custom rule, this test would only find the 6 errors that are caught by the standard ruleset.)

bok_choy/a11y/axe_core_ruleset.py
benpatterson: Ha. Looks like I missed it. :\

clytwynec: It will be more obvious in the read-the-docs rendering I think.

tests/test_accessibility.py
```diff
@@ -312,6 +317,41 @@ def test_check_for_accessibility_errors(self):
...
+        self.page.a11y_audit.config.customize_ruleset('tests/a11y_custom_rules.js')
+        self._do_audit_and_check_errors(7)
+
+    def test_customize_ruleset_bad_file_path(self):
```
clytwynec: @benpatterson 
re: "Could you add a test(s) for when the file is not found?"

I just added this test.

clytwynec:
@benpatterson @jzoldak  For a reference point, I've got PR https://github.com/edx/edx-platform/pull/10116 started for where I see this going.

benpatterson:
:+1:

clytwynec:
@jzoldak Could you review this when you get a chance?

bok_choy/a11y/axe_core_ruleset.py
```diff
@@ -148,6 +151,53 @@ def set_scope(self, include=None, exclude=None):
...
+        custom_file = custom_ruleset_file or os.environ.get(
+            "BOKCHOY_A11Y_CUSTOM_RULES_FILE"
+        )
+
+        if not custom_file:
+            return
+
+        with open(custom_file, "r") as additional_rules:
+            custom_rules = additional_rules.read()
+        self.custom_rules = custom_rules
```
jzoldak: Would it make sense to validate the format of the content of the rules file? E.g. that it starts with var customRules =  and it's valid JSON?
How does it fail if the file contents are not formatted correctly?

clytwynec: It's not actually JSON, it's javascript.  I think ensuring that it starts with "var customRules" makes sense, but think it makes more sense to leave validation of whether or not the rules are written properly to whoever is writing them.  (e.g. I've written javascript tests for the edx-specific rules I've been working on.)

jzoldak: awesome, yeah I'm good with at least ensuring that it starts with "var customRules = "

jzoldak: what does happen if it's not valid JS anyhow?

clytwynec: It would time out waiting for the audit results.

clytwynec: @jzoldak I've updated with this change.

jzoldak:
:+1:

PR #120, opened 2015-10-22
add BrowserQuery is_focused method and focused property

clytwynec:
@benpatterson @jzoldak  Please review.  I'm hoping to merge this and add it to the 0.4.7 release.  I know I've already seen (and implemented) a few workarounds for the lack of an `is_focused` method.  Since we're starting to test for focus issues more regularly, it would be really useful to have a standard way of doing it that is in line with the other patterns in bok choy.

### DIFF
--- a/bok_choy/query.py
+++ b/bok_choy/query.py
@@ -445,6 +445,25 @@ class BrowserQuery(Query):
         """
         return self.present and not self.visible
 
+    def is_focused(self):
+        """
+        Checks that _at least one_ matched element is focused. More
+        specifically, it checks whether the element is document.activeElement.
+        If no matching element is focused, this returns `False`.
+
+        Returns:
+            bool
+        """
+        active_el = self.browser.execute_script("return document.activeElement")
+        query_results = self.map(lambda el: el == active_el, 'focused').results
+
+        if query_results:
+            return any(query_results)
+        else:
+            return False
+
+    focused = property(is_focused)
+
     def click(self):
         """
         Click each matched element.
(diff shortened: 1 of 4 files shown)


bok_choy/query.py
```diff
@@ -445,6 +445,24 @@ def invisible(self):
         """
         return self.present and not self.visible
 
+    def is_focused(self):
+        """
+        Checks that at least one matched element is focused. More
+        specifically, if checks whether the element is document.activeElement.
+
+        Returns:
+            bool
```
jzoldak: This will return False if no element matches the browser query, right?
We should make that clear in the docstring, because I know that people had questions about that case when using the visible() method.

I think either with a code example or further explanation in the Returns section would be fine.

tests/test_focused.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    def setUp(self):
+        super(FocusedTest, self).setUp()
+        self.page = FocusedPage(self.browser).visit()
+
+    def test_focused(self):
+        def el_focused():
+            return self.page.q(css="#main-content").focused
+
+        self.assertFalse(el_focused())
```
jzoldak: Can you throw in a self.assertFalse(self.page.q(css="#i-dont-exist").focused) so that the test code also has an example showing the return value is False if the element does not exist?

clytwynec:
@jzoldak  I've updated to address your comments.

jzoldak:
:+1:

benpatterson:
:+1:

PR #121, opened 2015-11-03
Improve check for custom rules file format

clytwynec:
@benpatterson @jzoldak  Please review.  

Checking that the custom rules file started with "var customRules" was too strict, preventing even having comments at the top of the file.  Instead, this checks that "var customRules" is somewhere in the file, but not necessarily the first thing in it.


benpatterson:
:+1:

jzoldak:
:+1:

PR #122, opened 2015-12-12
Capability for custom firefox path.

benpatterson:


benpatterson:
@jzoldak @clytwynec 

This will allow us to use an environment var for setting a custom firefox path. This can help us more-smoothly upgrade our test infrastructure, by allowing older pull requests to use the older version of firefox, while we upgrade the rest of platform (e.g., we can install two firefox binaries on the same machine, and point to a particular installation with the env var).

Also FYI @raeeschachar @adampalay

bok_choy/browser.py
```diff
@@ -230,6 +231,12 @@ def _local_browser_class(browser_name):
             browser_kwargs = {
                 'firefox_profile': firefox_profile,
             }
+            if os.environ.get('SELENIUM_FIREFOX_PATH', None):
+                binary_kwarg = {
```
clytwynec: Is there a benefit to creating this as a separate dict and then updating the other one over just assigning the value directly?

(e.g. `browser_kwargs['firefox_binary'] = FirefoxBinary(firefox_path=os.environ.get('SELENIUM_FIREFOX_PATH'))`)

clytwynec:
@benpatterson Can you explain a bit more about how this will work in context?  I'm not sure I understand at what point this variable will be getting set.

bok_choy/browser.py
benpatterson: No particular reason. I can give that a look...sounds simpler.

benpatterson:
Hi @clytwynec in general what I'm thinking is, rather than telling folks to "rebase now" when we upgrade firefox (which will require a lock-step selenium library upgrade), it would be nice to have a way to fall back on the older browser. 

So, a jenkins-worker would have two firefoxes available. Old code (with old selenium library) would run by default on firefox 28. New code (with new selenium library) would set this env var and use a newer firefox version as a result. 

In general, this can give us a sort of supported window, giving folks more time to rebase. Then (for example) after a couple of weeks, we roll out workers with only the new firefox and remove the flag at that time.

Does that make more sense?

clytwynec:
Ok.  Sounds good.  Thanks for the explanation.   :+1:

adampalay:
:+1:

PR #124, opened 2015-12-12
Use docker for travis builds.

benpatterson:


benpatterson:
@jzoldak @clytwynec

scripts/install_browsermob_proxy.sh
```diff
@@ -9,27 +18,29 @@ BMP_URL="https://github.com/lightbody/browsermob-proxy/releases/download/$BMP_NA
...
+        unzip $TEMP_BROWSERMOB_DIR/bmp.zip -d $TEMP_BROWSERMOB_DIR
+        mv $TEMP_BROWSERMOB_DIR/$BMP_NAME $BROWSERMOB_INSTALL_LOC
     else
         echo "browsermob-proxy package found."
     fi
 
     echo "Finishing installation of browsermob-proxy."
     # make sure the packages main script is executable
-    chmod 0755 /etc/browsermob-proxy/bin/browsermob-proxy
+    chmod 0755 $TEMP_BROWSERMOB_DIR/browsermob-proxy/bin
```
clytwynec: Will this file be left in a tmp dir?

benpatterson: Ah good point. I've updated it with the latest commit. It will be the install location.

.travis.yml
```diff
@@ -12,9 +13,10 @@ install:
     - "pip install -r test_requirements.txt"
     - "pip install -r requirements.txt"
     - "pip install -r doc_requirements.txt"
-    - "sudo ./scripts/install_browsermob_proxy.sh"
+    - "./scripts/install_browsermob_proxy.sh"
 script:
     - "pylint bok_choy"
+    - "export PATH=browsermobtemp/browsermob-proxy/bin:$PATH"
```
clytwynec: Should this be using the install location too?

benpatterson: For travis, the location is one and the same. Saying it with variables (that are not exposed to travis, so we hardcode them), the path to the bin dir is `$TEMP_BROWSERMOB_DIR/browsermob-proxy/bin`, which is the same as `$BROWSERMOB_INSTALL_LOC/bin`

For the travis script, we use the path relative to project root.

clytwynec:
:+1:

PR #125, opened 2016-01-19
Update to version v0.4.8.

benpatterson:
@jzoldak @clytwynec 

re: TE-1136

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -31,16 +31,16 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'bok-choy'
-copyright = u'2015, EdX'
+copyright = u'2016, EdX'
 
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -31,16 +31,16 @@
 
 # General information about the project.
 project = u'bok-choy'
-copyright = u'2015, EdX'
+copyright = u'2016, EdX'
```
clytwynec: I don't really know enough about copyright ... are we actually supposed to change this each year?

benpatterson: Yeah, moreso for good hygiene as I understand it.

jzoldak:
:+1:

clytwynec:
:+1:

PR #128, opened 2016-01-21
v0.4.10 release. Contains no code changes.

benpatterson:


benpatterson:
FYI @jzoldak 

There are some sequences of events that need to take place, it appears. I'll merge this and see if it publishes.

PR #129, opened 2016-01-21
Disable whitelist in attempt to have a successful pypi deploy.

benpatterson:


benpatterson:
yeah. that didn't work.

PR #130, opened 2016-01-21
Remove pypi deploy via Travis.

benpatterson:
Doing so requires configuration that is not compatible with our workflow. Comments
stored in TE-1045.


benpatterson:
@jzoldak

jzoldak:
:+1: :rocket:

PR #133, opened 2016-01-23
Capability for scrolling to an element.

benpatterson:
Sometimes an element is not clickable because it is not in the viewport. However,
tests like `<query_object>.present` or `<query_object>.visible` will return
`True`.This change provides the capability for scrolling to an element in the
viewport.

### DIFF
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -564,3 +564,28 @@ class PageObject(object):
 
         """
         self.wait_for(lambda: self.q(css=element_selector).invisible, description=description, timeout=timeout)
+
+    @unguarded
+    def scroll_to_element(self, element_selector, timeout=60):
+        """
+        Scrolls browser such that the element specified appears at the top. Before scrolling, waits for
+        the element to be present.
+
+        Example usage:
+
+        .. code:: python
+
+            self.scroll_to_element('.far-down', 'Scroll to far-down')
+
+        Arguments:
+            element_selector (str): css selector of the element.
+            timeout (float): Maximum number of seconds to wait for the Promise to be satisfied before timing out
+
+        """
+        # Ensure element exists
+        msg = "Element '{element}' is present".format(element=element_selector)
+        self.wait_for(lambda: self.q(css=element_selector).present, msg, timeout=timeout)
+
+        # Obtain coordinates and use those for JavaScript call
+        loc = self.q(css=element_selector).first.results[0].location
+        self.browser.execute_script("window.scrollTo({x},{y})".format(x=loc['x'], y=loc['y']))
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -568,7 +568,7 @@ class PageObject(object):
     @unguarded
     def scroll_to_element(self, element_selector, timeout=60):
         """
-        Scrolls browser such that the element specified appears at the top. Before scrolling, waits for
+        Scrolls the browser such that the element specified appears at the top. Before scrolling, waits for
         the element to be present.
 
         Example usage:
@@ -579,7 +579,10 @@ class PageObject(object):
 
         Arguments:
             element_selector (str): css selector of the element.
-            timeout (float): Maximum number of seconds to wait for the Promise to be satisfied before timing out
+            timeout (float): Maximum number of seconds to wait for the element to be present on the
+                page before timing out.
+
+        Raises: BrokenPromise if the element does not exist (and therefore scrolling to it is not possible)
 
         """
         # Ensure element exists
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -14,7 +14,7 @@ function echo_task {
 }
 
 REPO_ROOT=`dirname $BASH_SOURCE`
-SERVER_PORT=8003
+SERVER_PORT=8005
 
 export BOK_CHOY_HAR_DIR=$REPO_ROOT/hars
 export LOG_DIR=$REPO_ROOT/logs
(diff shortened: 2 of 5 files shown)


run_tests.sh
```diff
@@ -14,7 +14,7 @@ function echo_task {
 }
 
 REPO_ROOT=`dirname $BASH_SOURCE`
-SERVER_PORT=8003
+SERVER_PORT=8005
```
benpatterson: author's note. I changed this to 8005 so it wouldn't interfere with a running edx-platform devstack.

benpatterson:
@jzoldak @clytwynec @raeeschachar

bok_choy/page_object.py
```diff
@@ -564,3 +564,28 @@ def wait_for_element_invisibility(self, element_selector, description, timeout=6
...
+
+            self.scroll_to_element('.far-down', 'Scroll to far-down')
+
+        Arguments:
+            element_selector (str): css selector of the element.
+            timeout (float): Maximum number of seconds to wait for the Promise to be satisfied before timing out
+
+        """
+        # Ensure element exists
+        msg = "wait for element '{element}' to be present".format(element=element_selector)
```
clytwynec: I think the pattern we have been using for descriptions in the `wait_for` calls is to describe the state that is being waited for.  (One of the examples in this file is `Submit Button is Present`.)  The output on failure then says "Broken Promise: Submit Button is Present".

tests/site/long_page.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Long Page</title>
+    <script src="//code.jquery.com/jquery-1.10.2.min.js"></script>
+    <script>
+        $(document).ready(function() {
```
clytwynec: Is this script needed?

tests/site/long_page.html
```diff
@@ -0,0 +1,33 @@
...
+            });
+        });
+    </script>
+
+</head>
+<body>
+    <p id="long_part"></p>
+    <p id="element_after_long_part">scroll here</p>
+    <p id="next_long_part"></p>
+    <script>
```
clytwynec: Is there a reason these scripts are here instead of in the head element?
It might be a more straightforward to set the heights of long parts directly.

tests/site/long_page.html
benpatterson: Och you are right. That was an earlier iteration on my part.

tests/site/long_page.html
benpatterson: No particular reason. Perhaps ignorant of me as far as javascript goes, but can you have scripts in HEAD that reference IDs in the body?

bok_choy/page_object.py
benpatterson: Good point. Updated.

tests/site/long_page.html
clytwynec: Yup, that's normal.  I think you'll need to continue use `$(document).ready` though, to make sure the element is loaded before the line that references it is run.

run_tests.sh
jzoldak: good call. thanks!

benpatterson:
@clytwynec @jzoldak @raeeschachar 

Updated based on comments.

tests/test_selector.py
```diff
@@ -49,3 +51,33 @@ def test_filtered_query(self):
...
+        super(ScrollTest, self).setUp()
+
+        self.long_page = LongPage(self.browser)
+        self.long_page.visit()
+
+    def test_scroll(self):
+        self.assertEquals(self._get_window_position(), 0)
+        self.long_page.scroll_to_element('#element_after_long_part')
+        # Different browsers, CI systems, and resolutions may present the
+        # element in varying locations. Use a greater-than instead of an equals.
```
clytwynec: One other suggestion would be to get the actual position of the element for the comparison. You can do this with jQuery using `offset`:

```
$('#element_after_long_part').offset().top
```

clytwynec:
One other suggestion for the tests. Up to you if you want to use it. 

:+1:

bok_choy/page_object.py
```diff
@@ -564,3 +564,28 @@ def wait_for_element_invisibility(self, element_selector, description, timeout=6
 
         """
         self.wait_for(lambda: self.q(css=element_selector).invisible, description=description, timeout=timeout)
+
+    @unguarded
+    def scroll_to_element(self, element_selector, timeout=60):
+        """
+        Scrolls browser such that the element specified appears at the top. Before scrolling, waits for
```
jzoldak: nit: Scrolls the browser

bok_choy/page_object.py
```diff
@@ -564,3 +564,28 @@ def wait_for_element_invisibility(self, element_selector, description, timeout=6
...
+
+        Example usage:
+
+        .. code:: python
+
+            self.scroll_to_element('.far-down', 'Scroll to far-down')
+
+        Arguments:
+            element_selector (str): css selector of the element.
+            timeout (float): Maximum number of seconds to wait for the Promise to be satisfied before timing out
```
jzoldak: nit: this assumes knowledge of what a Promise is. Maybe instead say ..to wait for the element to be present on the page.

jzoldak: nit: oh and maybe add in a "Raises" section to the docstring. I'm being picky here because I find myself often referring people to the bok choy readthedocs documentation which is built off the docstrings.

jzoldak:
:+1:

PR #135, opened 2016-02-02
WIP: Try running with this change.

cahrens:


bok_choy/page_object.py
```diff
@@ -282,11 +282,20 @@ def visit(self):
...
+            page_wait = self.wait_for_page()
         except BrokenPromise:
             raise PageLoadError("Timed out waiting to load page '{!r}' at URL '{}'".format(
                 self, self.url
             ))
+        
+        assert False, "Making sure this will fail"
+        assert "<XSS" not in self.browser.page_source, "page contains XSS vulnerability"
+        assert "<xss" not in self.browser.page_source, "page contains XSS vulnerability"
+
```
cgoldberg: both case sensitive cases can be handled in one:
`assert "<xss" not in self.browser.page_source.lower(), "page contains XSS vulnerability"`

cahrens: Thanks for the suggestion @cgoldberg. I'm just playing around with some ideas for automating xss testing-- don't worry, if this ever makes it to an official pull request state it will look nothing like this!

PR #119, opened 2015-10-21
v0.4.7

clytwynec:
@benpatterson @jzoldak

PR #139, opened 2016-03-11
Add xss vulnerability testing.

cahrens:
This PR introduces a framework for doing fairly automatic XSS testing (I say "fairly" because the test code is still responsible for injecting a specific string into appropriate places). For background, see:

https://openedx.atlassian.net/browse/TNL-4107

Examples of using this framework can be seen in this PR:
https://github.com/edx/edx-platform/pull/11758
Example failures with the flag enabled: https://build.testeng.edx.org/job/edx-platform-bok-choy-pr/15369/

@jzoldak and @robrap please review. FYI @benpatterson

### DIFF
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -10,6 +10,7 @@ import logging
 import os
 import socket
 import urlparse
+import re
 from textwrap import dedent
 from lazy import lazy
 
@@ -18,6 +19,15 @@ from selenium.common.exceptions import WebDriverException
 from .query import BrowserQuery
 from .promise import Promise, EmptyPromise, BrokenPromise
 from .a11y import AxeCoreAudit, AxsAudit
+
+
+# String that can be used to test for XSS vulnerabilities.
+# Taken from https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
+XSS_INJECTION = "'';!--\"<XSS>=&{()}"
+
+EXPECTED_INPUT_VALUE_FORMAT = re.compile(r'value="\'\';!--.*<xss.*{\(\)}"')
+
+XSS_HTML = "<xss"
 
 
 class WrongPageError(Exception):
@@ -30,6 +40,13 @@ class WrongPageError(Exception):
 class PageLoadError(Exception):
     """
     An error occurred while loading the page.
+    """
+    pass
+
+
+class XSSExposureError(Exception):
+    """
+    An XSS issue has been found on the current page.
     """
     pass
 
@@ -179,6 +196,8 @@ class PageObject(object):
         self.browser = browser
         a11y_flag = os.environ.get('VERIFY_ACCESSIBILITY', 'False')
         self.verify_accessibility = a11y_flag.lower() == 'true'
+        xss_flag = os.environ.get('VERIFY_XSS', 'False')
+        self.verify_xss = xss_flag.lower() == 'true'
 
     @lazy
     def a11y_audit(self):
@@ -330,6 +349,30 @@ class PageObject(object):
             )
             raise WrongPageError(msg)
 
+    def _verify_xss_exposure(self):
+        """
+        Verify that there are no obvious XSS exposures on the page (based on test authors
+        including XSS_INJECTION in content rendered on the page).
+
+        If an xss issue is found, raise a 'XSSExposureError'.
+        """
+        # Use innerHTML to get dynamically injected HTML as well as server-side HTML.
+        html_source = self.browser.execute_script(
+            "return document.documentElement.innerHTML.toLowerCase()"
+        )
+
+        # Check taken from https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
+        all_hits_count = html_source.count(XSS_HTML)
+        if all_hits_count > 0:
+            safe_hits_count = len(EXPECTED_INPUT_VALUE_FORMAT.findall(html_source))
+            if all_hits_count > safe_hits_count:
+                potential_hits = re.findall('<[^<]+<xss', html_source)
+                raise XSSExposureError(
+                    "{} XSS issue(s) found on page. Potential places are {}".format(
+                        all_hits_count - safe_hits_count, potential_hits
+                    )
+                )
+
     @unguarded
     def wait_for_page(self, timeout=30):
         """
@@ -385,6 +428,8 @@ class PageObject(object):
         Returns:
             BrowserQuery
         """
+        if self.verify_xss:
+            self._verify_xss_exposure()
         return BrowserQuery(self.browser, **kwargs)
 
     @contextmanager
--- a/docs/xss.rst
+++ b/docs/xss.rst
@@ -1,0 +1,100 @@
+Performing XSS Vulnerability Audits
+===================================
+
+The bok-choy framework includes the ability to perform XSS (cross-site scripting) audits on
+web pages using a short XSS locator defined in
+https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
+
+Trigger XSS Vulnerability Audits in Existing Tests
+--------------------------------------------------
+
+You might already have some bok-choy tests written for your web application. To
+leverage existing bok-choy tests and have them fail on finding XSS vulnerabilities,
+follow these steps.
+
+1. Insert the ``XSS_INJECTION`` string defined in ``bok_choy.page_object`` into your page content.
+2. Set the ``VERIFY_XSS`` environment variable to ``True``.
+
+::
+
+    export VERIFY_XSS=True
+
+
+With this environment variable set, an XSS audit is triggered whenever a page object's ``q``
+method is called. The audit will detect improper escaping both in HTML and in Javascript
+that is embedded within HTML.
+
+If errors are found on the page, an XSSExposureError is raised.
+
+Here is an example of a bok-choy test that will check for XSS vulnerabilities.
+It clicks a button on the page, and the user's name is inserted into the page.
+If the user name is not properly escaped, the display
+of the name (which is data provided by the user and thus potentially malicious) can cause
+XSS issues.
+
+In the case of the ``test_button_click_output`` test case in the example below,
+an audit will be done in the ``click_button()``, ``output()``, and ``visit()`` method calls,
+as each of those will call out to ``q``.
+
+If any XSS errors are found, then the test case will fail with an
+XSSExposureError.
+
+.. code-block:: python
+
+    from bok_choy.page_object import PageObject, XSS_INJECTION
+
+
+    class MyPage(PageObject):
+        def url(self):
+            return 'https://www.mysite.com/page'
+
+        def is_browser_on_page(self):
+            return self.q(css='div#fixture button').present
+
+        def click_button(self):
+            """
+            Click on the button element (id="button").
+            On my example page this will trigger an ajax call
+            that updates the #output div with the user's name.
+            """
+            self.q(css='div#fixture button').first.click()
+            self.wait_for_ajax()
+
+        @property
+        def output(self):
+            """
+            Return the contents of the "#output" div on the page.
+
+            In the example page, it will contain the user's name after being
+            updated by the ajax call that is triggered by clicking the button.
+            """
+            text_list = self.q(css='#output').text
+
+            if len(text_list) < 1:
+                return None
+            else:
+                return text_list[0]
+
+    class MyPageTest(WebAppTest):
+        def setUp(self):
+            """
+            Log in as a particular user.
+            """
+            super(MyPageTest, self).setUp()
+            self.user_name = XSS_INJECTION
+            self.log_in_as_user(self.user_name)
+
+        def test_button_click_output(self):
+            page = MyPage(self.browser)
+            page.visit()
+            page.click_button()
+
+            self.assertEqual(page.output, self.user_name)
+
+        def log_in_as_user(self, user_name):
+            """
+            Would be implemented to log in as a particular user
+            with a potentially malicious, user-provided name.
+            """
+            pass
+
--- a/tests/site/xss_safe.html
+++ b/tests/site/xss_safe.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>XSS Safe</title>
+</head>
+<body>
+    <script>
+        window.escaped = "\u0027\u0027\u003B!\u002D\u002D\u0022\u003CXSS\u003E\u003D\u0026{()}";
+    </script>
+    <div id="fixture">
+        <div class="escaped">&#39;&#39;;!--&#34;&lt;XSS&gt;=&amp;{()}</div>
+        <input type="text" value="&#39;&#39;;!--&#34;&lt;XSS&gt;=&amp;{()}">
+    </div>
+</body>
+</html>
--- a/tests/test_xss_exposure.py
+++ b/tests/test_xss_exposure.py
@@ -1,0 +1,53 @@
+"""
+Tests for identifying XSS vulnerabilities.
+This is currently done when the "q" method is called.
+"""
+
+import os
+from mock import patch
+
+from bok_choy.web_app_test import WebAppTest
+from .pages import SitePage
+from bok_choy.page_object import XSSExposureError
+
+
+class XSSExposureTest(WebAppTest):
+    """
+    Tests for identifying XSS vulnerabilities.
+    """
+    def _visit_page(self, page_name):
+        self.site_page = SitePage(self.browser)
+        self.site_page.name = page_name
+        self.site_page.visit()
+
+    @patch.dict(os.environ, {'VERIFY_XSS': 'True'})
+    def test_html_exposure(self):
+        self._visit_page("xss_html")
+        with self.assertRaisesRegexp(XSSExposureError, "2 XSS issue"):
+            self.site_page.q(css='.unescaped')
+
+    @patch.dict(os.environ, {'VERIFY_XSS': 'True'})
+    def test_js_exposure(self):
+        self._visit_page("xss_js")
+        with self.assertRaisesRegexp(XSSExposureError, "1 XSS issue"):
+            self.site_page.q(css='.unescaped')
+
+    @patch.dict(os.environ, {'VERIFY_XSS': 'True'})
+    def test_mixed_exposure(self):
+        self._visit_page("xss_mixed")
+        with self.assertRaisesRegexp(XSSExposureError, "2 XSS issue"):
+            self.site_page.q(css='.unescaped')
+
+    @patch.dict(os.environ, {'VERIFY_XSS': 'True'})
+    def test_escaped(self):
+        self._visit_page("xss_safe")
+        self.site_page.q(css='.escaped')
+
+    @patch.dict(os.environ, {'VERIFY_XSS': 'False'})
+    def test_xss_testing_disabled_explicitly(self):
+        self._visit_page("xss_html")
+        self.site_page.q(css='.unescaped')
+
+    def test_xss_testing_disabled_by_default(self):
+        self._visit_page("xss_html")
+        self.site_page.q(css='.unescaped')
--- a/tests/site/xss_html.html
+++ b/tests/site/xss_html.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>XSS HTML</title>
+</head>
+<body>
+    <div id="fixture">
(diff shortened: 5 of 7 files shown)


docs/xss.rst
```diff
@@ -0,0 +1,95 @@
+Performing XSS Vulnerability Audits
```
cahrens: @catong can you please help me by reviewing this doc? It's pretty minimal at the moment (and the example case is a bit contrived-- you can see more realistic usages in https://github.com/edx/edx-platform/pull/11758).

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+==================================
+
+The bok-choy framework includes the ability to perform XSS (cross-site scripting) audits on
+web pages using a short XSS locator defined in
+https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
+
+Leverage Your Existing Tests and Fail on XSS Vulnerability Errors
+-------------------------------------------------------------
```
catong: Not sure if this matters here for docs that just live in a repo, but typically in headings make sure the character formatting (===, -----, whatever) equals or exceeds the number of characters in the heading.

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+
```
catong: Suggestions: 
1. Modify the heading in line 8 to something like "Trigger XSS Vulnerability Audits in Existing Tests".
2. Move your first sentence from line 19 to create an intro after the heading. "You might already have some bok-choy tests written for your web application. To leverage existing bok-choy tests and have them fail on finding XSS vulnerabilities, follow these steps." 
3. Add a set of numbered steps after your intro sentence.
4. Move the prereq in your note(line 16-17) into step 1. 

Steps would look something like this:
1. "Inject the XSS_INJECTION string defined in... into your page content." 
2. "Set the VERIFY_XSS environment variable ..."
    (Move code example here: ```export VERIFY_XSS = True```)

    An XSS audit is triggered whenever a page object's ```q``` method is called.
    If errors are found on the page, an XSS ExposureError is raised.

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+To trigger XSS vulnerability audits passively, set the ``VERIFY_XSS``
+environment variable to ``True``. Doing so triggers an XSS audit
+whenever a page object's ``q`` method is called. If errors are
+found on the page, an XSSExposureError is raised.
+
+.. note:: XSS vulnerabilities will only be found if you have injected the specific string
+XSS_INJECTION defined in bok_choy.page_object into your page content.
+
+You might already have some bok-choy tests written for your web application.
```
catong: Move this sentence earlier, to introduce the numbered steps.

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+Here is an example of a bok-choy test that will implicitly check for XSS vulnerabilities.
+It involves clicking on a button on page, and it is assumed that the user's name is
```
catong: "It involves clicking a button to display the user's name on the page." (Is this still accurate?)

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+            self.assertEqual(page.output, 'yes!')
+
+        def log_in_user(self, user_name):
+            """
+            Would be implemented to log in as a particular user
+            with a potentially malicious, user-provided name.
+            """
+            pass
+
+Before running your bok-choy tests, set the environment variable
```
catong: I'm not sure if this was meant to be in the code example, or a comment after it? I don't think you need it because Step 2 above is where you're telling people to set the env var. You can include the code example with that step.

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+``VERIFY_XSS`` to ``True``.
+
+::
+
+    export VERIFY_XSS=True
+
+This will trigger an audit, looking for unescaped usage of XSS_INJECTION, whenever a call to
+``q`` is made. The audit will detect improper escaping both in HTML and JavaScipt embedded
+within HTML.
```
catong: This information can be removed here; I would put it in the result of Step 2.

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+
+In the case of the ``test_button_click_output`` test case in the example above,
```
catong: Possibly -- move the info in lines 90 -95 to BEFORE the test example, (changing the reference to "example below") and just end this document with the end of the code example.

"In the example below, an audit is done in the ... method calls, as each of those will call out to ``q``. The test case fails with an XSSExposureError if any XSS errors are found."

catong:
@cahrens I made my suggestions in inline comments. Let me know if it doesn't make sense, or if a verbal walkthrough would be helpful!

benpatterson:
Just starting a look at this and I had a couple comments...
* To @catong as well, we use readthedocs for our bok-choy documentation, and it might be better to move the documentation over to there. Honestly I haven't looked at the tutorial in awhile, so I can't remember if that's also in docstrings, but if you look at browser.py you can see some large docstrings, which is because we use readthedocs. So, if you haven't already, I'd look at getting sphinx to take care of some of the packaging.
* As noted on other PRs, the CHANGELOG has several things that would be needed for the next release. I'd rather the setup.py, CHANGELOG, and other release-related changes be a separate PR (or something that JZ or I take care of after this PR gets merged).

I know @jzoldak  may have comments here, but I'll also take a look at the code for a pass. Thanks a lot for this PR.

bok_choy/page_object.py
```diff
@@ -330,6 +345,25 @@ def _verify_page(self):
...
+        html_source = self.browser.execute_script(
+            "return document.documentElement.innerHTML.toLowerCase()"
+        )
+
+        # Check taken from https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
+        all_hits = html_source.count('<xss')
+        if all_hits > 0:
+            value_hits = len(re.findall(r'value="\'\';!--.*<xss.*{\(\)}"', html_source))
+            if all_hits > value_hits:
+                raise XSSExposureError("{} XSS issue(s) found on page".format(all_hits - value_hits))
```
benpatterson: would it be good to have a debug log with the html source, since it's created & lower-cased above? Just thinking of helping troubleshoot if an issue is found.

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+The bok-choy framework includes the ability to perform XSS (cross-site scripting) audits on
+web pages using a short XSS locator defined in
+https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
+
+Leverage Your Existing Tests and Fail on XSS Vulnerability Errors
+-------------------------------------------------------------
+
+To trigger XSS vulnerability audits passively, set the ``VERIFY_XSS``
+environment variable to ``True``. Doing so triggers an XSS audit
+whenever a page object's ``q`` method is called. If errors are
```
benpatterson: I think this refers to the proof-of-concept implementation...since it's now on `wait_for_page`

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+.. code-block:: python
+
+    from bok_choy.page_object import PageObject, XSS_INJECTION
+
+
+    class MyPage(PageObject):
+        def url(self):
+            return 'https://www.mysite.com/page'
+
+        def click_button(self):
```
benpatterson: nit: for sake of completion (since `url` is defined), `is_browser_on_page` should be defined.

docs/xss.rst
cahrens: @benpatterson It's back to q because that finds more failures (the other option was verify_page, not wait_for_page). For details, see comments at end of https://openedx.atlassian.net/wiki/display/TE/Design%3A+Add+XSS+Vulnerability+Testing+feature+to+Bok+Choy.

docs/xss.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Before running your bok-choy tests, set the environment variable
+``VERIFY_XSS`` to ``True``.
+
+::
+
+    export VERIFY_XSS=True
+
+This will trigger an audit, looking for unescaped usage of XSS_INJECTION, whenever a call to
+``q`` is made. The audit will detect improper escaping both in HTML and JavaScipt embedded
```
benpatterson: same here, and i'm wondering if the example would be updated for the `wait_for_page` approach.

bok_choy/page_object.py
cahrens: That's a good idea. Do you or @jzoldak have a suggest for the best way to log that information (which will likely be very long)?

benpatterson:
Finished my first pass...I had comments about updating docs to use `wait_for_page` but it might make more sense to hear @jzoldak's opinion on the right place to set the xss check and then update code+doc appropriately. 

Might also be worth pointing out how much of a penalty (if any) for enabling the check.

cahrens:
Thanks, @benpatterson. As I mentioned (I imagine you just haven't seen the comment yet), @jzoldak and I have already discussed keeping the check in q vs. verify_page. 

It's a good idea to include performance information. I'll have to add that though after we do the profiling story.

benpatterson:
Oh great...if you guys have already figured it out, that sgtm.

I'll catch up on the convo...

sent via Android

tests/site/xss_safe.html
```diff
@@ -0,0 +1,15 @@
...
+<head>
+    <title>XSS Safe</title>
+</head>
+<body>
+    <script>
+        window.escaped = "\u0027\u0027\u003B!\u002D\u002D\u0022\u003CXSS\u003E\u003D\u0026{()}";
+    </script>
+    <div id="fixture">
+        <div class="escaped">&#39;&#39;;!--&#34;&lt;XSS&gt;=&amp;{()}</div>
+        <input type="text" value="'';!--\&quot;<XSS>=&amp;{()}">
```
cahrens: @robrap See this test case where some things are escaped, but <XSS> is not.

robrap: @dianakhuang @efischer19 @cahrens Here's what I did:
1. Saved a local file with the contents of the HTML above, except I fully escape the contents of the value attribute (i.e. the value will match the contents of the `<div>`).
2. When I load this file in my browser, the text box correctly contains the value we expect.
3. When I view source in the browser, it displays the source as it was (i.e. the contents of my original file).

My current theory is that this code is written to work around a defect in the Accounts Setting page.  I'm not sure if there is still a defect there, or if it was fixed by the change around the jQuery val() call.  

Can you do what I did above in this test and see if we really do get a false positive?

tests/site/xss_safe.html
```diff
@@ -9,7 +9,7 @@
     </script>
     <div id="fixture">
         <div class="escaped">&#39;&#39;;!--&#34;&lt;XSS&gt;=&amp;{()}</div>
-        <input type="text" value="'';!--\&quot;<XSS>=&amp;{()}">
+        <input type="text" value="&#39;&#39;;!--&#34;&lt;XSS&gt;=&amp;{()}">
```
efischer19: FYI @robrap

bok_choy/page_object.py
```diff
@@ -330,6 +345,25 @@ def _verify_page(self):
...
+        """
+        # Use innerHTML to get dynamically injected HTML as well as server-side HTML.
+        html_source = self.browser.execute_script(
+            "return document.documentElement.innerHTML.toLowerCase()"
+        )
+
+        # Check taken from https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
+        all_hits = html_source.count('<xss')
+        if all_hits > 0:
+            value_hits = len(re.findall(r'value="\'\';!--.*<xss.*{\(\)}"', html_source))
```
jzoldak: It would make the code more clear/understandable to have this regular expression pattern re.compile'd into a regular expression object that is saved into a global variable up next to XSS_INJECTION at the top of the file, as the two are interrelated.

I would do the same with the hardcoded '<xss' string 2 lines up (put it into a global variable up at the top).

docs/xss.rst
jzoldak: FWIW this doc doesn't just live in the repo, it gets processed by Sphinx and uploaded to readthedocs as the official documentation for the package. http://bok-choy.readthedocs.org/

tests/test_xss_exposure.py
```diff
@@ -0,0 +1,47 @@
+"""
+Tests for identifying XSS vulnerabilities.
+This is currently done with the "q" method is called.
```
jzoldak: typo /s/with/when/

tests/site/xss_safe.html
```diff
@@ -0,0 +1,15 @@
...
+<head>
+    <title>XSS Safe</title>
+</head>
+<body>
+    <script>
+        window.escaped = "\u0027\u0027\u003B!\u002D\u002D\u0022\u003CXSS\u003E\u003D\u0026{()}";
+    </script>
+    <div id="fixture">
+        <div class="escaped">&#39;&#39;;!--&#34;&lt;XSS&gt;=&amp;{()}</div>
+        <input type="text" value="&#39;&#39;;!--&#34;&lt;XSS&gt;=&amp;{()}">
```
robrap: @cahrens - @efischer19 determined that it wasn't that this page source shouldn't be escaped just as everything else is escaped.  The problem is that the `innerHtml` call off the document element outputs values with a different set of rules, so the regex needs to account for that.

Note that when we stopped the test on this page, and used View Page Source from the browser, we would again see this original fully escaped text.

bok_choy/page_object.py
jzoldak: Couple ideas:
* We might give #85 some love. That way the person could see the entire source of the rendered page. Not sure though if it's not helpful unless we use the document.documentElement.innerHTML trick from above.
* Something like below would at least narrow down to where they were on the page. This assumes (I’m not sure if it’s a valid assumption) that the vulnerability will be in the HTML after another occurrence of “<”. I think this might hold true because in HTML you’re at least past the body tag and in JS past a script tag.
```
        all_hits_count = html_source.count('<xss')
        if all_hits_count > 0:
            safe_hits = re.findall(r'value="\'\';!--.*<xss.*{\(\)}"', html_source)
            safe_hits_count = len(safe_hits)
            if all_hits_count > safe_hits_count:
                potential_hits = re.findall('<[^<]+<xss', html_source)
                raise XSSExposureError(
                    "{} XSS issue(s) found on page. Potential places are {}".format(all_hits_count - safe_hits_count, potential_hits)
                )
```

tests/test_xss_exposure.py
```diff
@@ -0,0 +1,47 @@
...
+            self.site_page.q(css='.unescaped')
+
+    @patch.dict(os.environ, {'VERIFY_XSS': 'True'})
+    def test_js_exposure(self):
+        self._visit_page("xss_js")
+        with self.assertRaisesRegexp(XSSExposureError, "1 XSS issue"):
+            self.site_page.q(css='.unescaped')
+
+    @patch.dict(os.environ, {'VERIFY_XSS': 'True'})
+    def test_escaped(self):
```
jzoldak: It would make sense to also add a testcase for a page that has mixed safe and unsafe areas.

bok_choy/page_object.py
cahrens: Great, I will make this code change.

dianakhuang:
@catong Could you take another look at the documentation? I made the fixes you suggested.

cahrens:
@jzoldak and/or @benpatterson can you review again? I believe your comments are all addressed.

catong:
@dianakhuang the 2 paras from lines 23-26 just need to be outdented so that they don't appear in code format. Otherwise :+1: !

setup.py
```diff
@@ -2,7 +2,7 @@
 
 from setuptools import setup
 
-VERSION = '0.4.11'
+VERSION = '0.5.0'
```
cahrens: @benpatterson or @jzoldak -- do you want this version change to be in a different PR? I believe there is an email thread about the next version.

tests/test_xss_exposure.py
cahrens: This has been added.

setup.py
jzoldak: Yes, please. For this repo we prefer to make the changes to setup.py, CHANGELOG, and docs/conf.py [all in a discreet PR like this example one](https://github.com/edx/bok-choy/pull/119). Mostly because otherwise at least one of them ends up being forgotten. :|

Let's get this PR merged in first, and then @benpatterson or I will release the new version, which is represented by TE-1218 and in our current sprint.

docs/xss.rst
```diff
@@ -0,0 +1,96 @@
...
+If errors are found on the page, an XSSExposureError is raised.
+
+Here is an example of a bok-choy test that will implicitly check for XSS vulnerabilities.
+It clicks a button on the page and assumes that the user's name is
+displayed on the page. If the user name is not properly escaped, the display
+of the name (which is data provided by the user and thus potentially malicious) can cause
+XSS issues.
+
+In the case of the ``test_button_click_output`` test case in the example below,
+an audit will be done in the ``click_button()`` ``output()`` and method calls,
```
jzoldak: typo, should be: ``click_button()`` and ``output()`` method calls

jzoldak: as well as the page.visit() call, because this in turn calls the page's is_browser_on_page method.

docs/xss.rst
```diff
@@ -0,0 +1,96 @@
...
+
+1. Insert the ``XSS_INJECTION`` string defined in ``bok_choy.page_object`` into your page content.
+2. Set the ``VERIFY_XSS`` environment variable to ``True``.
+
+::
+
+    export VERIFY_XSS=True
+
+
+An XSS audit is triggered whenever a page object's ``q`` method is called. The audit will detect
```
jzoldak: With this environment variable set, an XSS audit will be triggered whenever...

docs/xss.rst
```diff
@@ -0,0 +1,96 @@
...
+improper escaping both in HTML and in Javascript that is embedded within HTML.
+
+If errors are found on the page, an XSSExposureError is raised.
+
+Here is an example of a bok-choy test that will implicitly check for XSS vulnerabilities.
```
jzoldak: nit: I think the word "implicitly" is confusing here. It's not really implicit, as the test-writer intentionally imported XSS_INJECTION, and used it as the user_name. Maybe just leave it off.

dianakhuang:
@jzoldak Thanks for the doc comments. I believe I've addressed them all.

docs/xss.rst
```diff
@@ -0,0 +1,96 @@
...
+            that updates the #output div with the text "yes!"
+            """
+            self.q(css='div#fixture button').first.click()
+            self.wait_for_ajax()
+
+        @property
+        def output(self):
+            """
+            Return the contents of the "#output" div on the page.
+            """
```
jzoldak: nit: I had to read through the code twice to understand why the assertion was looking for the output to be 'yes!'.
Maybe it would have been more clear the first time through if this docstring also said something like:

> In the example page, it will contain the text "yes!" after being updated by the ajax call that is triggered by clicking the button.

This might be overkill, but I'd rather over-explain everything in the docs to account for new coders, non-native English speakers, etc.

docs/xss.rst
```diff
@@ -0,0 +1,100 @@
...
+            self.username, self.user_id = self.log_in_as_user(self.user_name)
+
+        def test_button_click_output(self):
+            page = MyPage(self.browser)
+            page.visit()
+            page.click_button()
+
+            self.assertEqual(page.output, 'yes!')
+
+        def log_in_user(self, user_name):
```
jzoldak: Is this a typo, should it be self. log_in_as_user?

For a new user of this feature, I'm concerned that they'll get hung up on the explanation in the text that says:

> It clicks a button on the page and assumes that the user's name is displayed on the page.

Because the test never actually checks that condition.

As long as this is a fictional app anyhow, maybe rather than putting "yes!" into an output div, the page could enter the user's name into the div?

docs/xss.rst
```diff
@@ -0,0 +1,100 @@
...
+                return text_list[0]
+
+    class MyPageTest(WebAppTest):
+        def setUp(self):
+            """
+            Log in as a particular user.
+            """
+            super(MyPageTest, self).setUp()
+            self.user_name = XSS_INJECTION
+            self.username, self.user_id = self.log_in_as_user(self.user_name)
```
jzoldak: nit: I'm hung up on the log_in_as_user method not returning anything. Maybe removing the `self.username, self.user_id =` bit will make the code seem more correct, even if it is just an example.

cahrens: I can see that, Jay. In my "real life" example, the method returns something. But for the purposes of this example test, it doesn't make sense.

@dianakhuang one more update?

jzoldak:
:+1:  :rocket:

PR #140, opened 2016-03-14
Update reference. We use slack now not IRC.

jzoldak:
@benpatterson or @jbarciauskas

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,7 +38,5 @@ Please do not report security issues in public. Please email security@edx.org
 Mailing List and IRC Channel
 ----------------------------
 
-You can discuss this code on the `edx-code Google Group`__ or in the
-``edx-code`` IRC channel on Freenode.
-
-__ https://groups.google.com/forum/#!forum/edx-code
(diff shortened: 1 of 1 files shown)


nedbat:
ugh, we should really have this info in one central place, and just point to it from each repo instead of having to tweak them all when we change our minds next time.

jzoldak:
@nedbat I totally agree. Not sure how to do that tho? Do you know where a good place to put that would be and how to pull in the info?

benpatterson:
I see the discussion....how about a wiki page?

jzoldak:
Oh, hm, maybe we just point them to https://open.edx.org/ ?

nedbat:
@jzoldak btw, I didn't mean to hold this up from getting merged.  It's good to fix the references we have.

benpatterson:
Yeah this is a :+1: from me. The centralized-info change could come later.

jzoldak:
Ok cool, merging this now. We'll figure out the other stuff later.

PR #141, opened 2016-03-17
Retry for a browser instantiation failure.

benpatterson:
Exceptions like "Can't load profile path" or "The browser appears
to have exited before we could connect" will now be handled with
a Promise and a try limit of 5.

### DIFF
--- a/bok_choy/browser.py
+++ b/bok_choy/browser.py
@@ -10,6 +10,7 @@ import socket
 from needle.driver import (NeedleFirefox, NeedleChrome, NeedleIe,
                            NeedleSafari, NeedlePhantomJS)
 from selenium import webdriver
+from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.chrome.options import Options
 from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
 
@@ -200,19 +201,20 @@ def browser(tags=None, proxy=None):
 
             return True, browser_class(*browser_args, **browser_kwargs)
 
-        except socket.error as err:
-            LOGGER.debug('Failed to instantiate browser: ' + err.strerror)
+        except (socket.error, WebDriverException) as err:
+            msg = str(err)
+            LOGGER.debug('Failed to instantiate browser: ' + msg)
             return False, None
 
     browser_instance = Promise(
-        browser_check_func, "Browser is instantiated successfully.", timeout=30).fulfill()
+        browser_check_func, "Browser is instantiated successfully.", try_limit=3).fulfill()
 
     return browser_instance
 
 
 def _local_browser_class(browser_name):
     """
-    Returns class, kwargs, and args needed to instatiate the local browser.
+    Returns class, kwargs, and args needed to instantiate the local browser.
     """
 
     # Log name of local browser
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -1,8 +1,16 @@
+"""
+Tests browser instantiation, selection, etc
+"""
+
 import tempfile
 import shutil
 import os
+import socket
+
 from selenium import webdriver
+from selenium.common.exceptions import WebDriverException
 import bok_choy.browser
+from bok_choy.promise import BrokenPromise
 from unittest import TestCase
 from mock import patch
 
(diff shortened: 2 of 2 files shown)


benpatterson:
@jzoldak @raeeschachar @Muddasser

bok_choy/browser.py
```diff
@@ -204,8 +205,13 @@ def browser_check_func():
             LOGGER.debug('Failed to instantiate browser: ' + err.strerror)
             return False, None
 
+        except WebDriverException as err:
+            msg = str(err)
```
jzoldak: What's the difference between str(err) and err.strerror (which is used in the other except block above)? Should we standardize on one or the other?

bok_choy/browser.py
```diff
@@ -204,8 +205,13 @@ def browser_check_func():
...
+            LOGGER.debug(msg)
+            return False, None
+
     browser_instance = Promise(
-        browser_check_func, "Browser is instantiated successfully.", timeout=30).fulfill()
+        browser_check_func, "Browser is instantiated successfully.", try_limit=5).fulfill()
```
jzoldak: Is it reasonable that this would fail 4 times but succeed the 5th? I'm thinking we might want to set the try_limit lower, maybe 2 or 3.

bok_choy/browser.py
benpatterson: Looks like err.strerror will bomb out if there is not a valid string raised in the exception. Fixing this to use the same pattern I was using.

bok_choy/browser.py
benpatterson: I went with 3 (in part b/c I don't know how often socket errors happen)

benpatterson:
@jzoldak I made some updates, added tests for both browser exception types, and improved a bit o' linting while I was there. Ready for 2nd pass.

bok_choy/browser.py
```diff
@@ -201,18 +202,24 @@ def browser_check_func():
             return True, browser_class(*browser_args, **browser_kwargs)
 
         except socket.error as err:
-            LOGGER.debug('Failed to instantiate browser: ' + err.strerror)
+            msg = str(err)
+            LOGGER.debug('Failed to instantiate browser: ' + msg)
+            return False, None
+
+        except WebDriverException as err:
```
jzoldak: I'm not sure it's helpful to the user to separate the two as all they would see is the msg anyhow (and "Failed to instantiate browser: " only in the case of a socket error but without seeing the code the user wouldn't know that.)

I think it's preferable to instead combine the two with:
`except (socket.error, WebDriverException) as err:`

tests/test_browser.py
```diff
@@ -95,3 +106,29 @@ def test_save_driver_logs(self):
...
             self.assertTrue(os.path.isfile(expected_file))
+
+    def test_profile_error(self):
+        """
+        If there is a WebDriverException when instantiating the driver,
+         it should be tried again.
+        """
+        patcher = patch('bok_choy.browser._local_browser_class')
+        patch_object = patcher.start()
+        patch_object.side_effect = WebDriverException()
```
jzoldak: A small improvement to the mock would be to have it raise the error with parameters.
From http://selenium-python.readthedocs.org/api.html#selenium.common.exceptions.WebDriverException maybe something like:
`patch_object.side_effect = WebDriverException(msg='that did not work', screen=None, stacktrace=None)`

tests/test_browser.py
```diff
@@ -95,3 +106,29 @@ def test_save_driver_logs(self):
...
+        self.assertEqual(patch_object.call_count, 3)
+
+    def test_socket_error(self):
+        """
+        If there is a WebDriverException when instantiating the driver,
+         it should be tried again.
+        """
+        patcher = patch('bok_choy.browser._local_browser_class')
+        patch_object = patcher.start()
+        patch_object.side_effect = socket.error
```
jzoldak: A small improvement to the mock would be to have it raise the error with parameters.
From https://docs.python.org/2/library/socket.html#socket.error maybe something like:
`patch_object.side_effect = socket.error(61, 'socket error message')`

tests/test_browser.py
benpatterson: done

tests/test_browser.py
benpatterson: done

bok_choy/browser.py
benpatterson: done.

benpatterson:
@jzoldak Thanks for the feedback; good suggestions. Updated.

jzoldak:
:+1:

benpatterson:
Updated a docstring, squashed, and now merging.

PR #144, opened 2016-03-25
Generalize the XSS false positives check.

cahrens:
While working on cleaning up a set of templates, I found that false positives were not limited to the value attribute of the input tag (see https://github.com/edx/edx-platform/pull/11926).

@robrap and I decided the best thing was just to allow the very specific escaping that we get from such an attribute field (as returned by the document method).

I verified that we do not lose any actual failures with this change (https://build.testeng.edx.org/view/edx-platform-specific-tests/job/edx-platform-bok-choy-custom/255/#showFailuresLink).

Ticket: https://openedx.atlassian.net/browse/TNL-4295

Please review @jzoldak and @robrap 
Note that I will want to release a new version of bok choy after this.

### DIFF
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -25,7 +25,11 @@ from .a11y import AxeCoreAudit, AxsAudit
 # Taken from https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
 XSS_INJECTION = "'';!--\"<XSS>=&{()}"
 
-EXPECTED_INPUT_VALUE_FORMAT = re.compile(r'value="\'\';!--.*<xss.*{\(\)}"')
+# When the injected string appears within an attribute (for instance, value of an input tag,
+# or alt of an img tag), if it is properly escaped this is the format we will see from
+# document.documentElement.innerHTML. To avoid false positives, we need to allow this
+# specific string, which hopefully is unique/odd enough that it would never appear accidentally.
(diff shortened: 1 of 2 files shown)


bok_choy/page_object.py
```diff
@@ -25,7 +25,7 @@
 # Taken from https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
 XSS_INJECTION = "'';!--\"<XSS>=&{()}"
 
-EXPECTED_INPUT_VALUE_FORMAT = re.compile(r'value="\'\';!--.*<xss.*{\(\)}"')
+EXPECTED_ATTRIBUTE_FORMAT = re.compile(r'\'\';!--&quot;<xss>=&amp;{\(\)}')
```
robrap: Does it make sense to just drop `value`?
```
EXPECTED_INPUT_VALUE_FORMAT = re.compile(r'="\'\';!--.*<xss.*{\(\)}"')
```

cahrens: No, because many of our common usages of the XSS injection string concatenate the string with some other part (for example, to make it unique). We would still get false positives for those. In fact, that was the reason why we had a false positive in the original 17 test failures, before this fix was made.

robrap:
:+1:  with a comment to explain why this is needed (i.e. that this output is not the same as one would see when viewing page source).

jzoldak:
:+1:

PR #146, opened 2016-04-11
Port to Python 3.5+ using six

dstufft:
Fixes #137

PR #147, opened 2016-04-13
Support Opera as a browser.

benpatterson:


benpatterson:
@jzoldak @nedbat 

and also @dstufft  (since Ned pointed out you ran into this issue)

dstufft:
Looks good to me.

For what it's worth there's also the Edge browser, but I don't know enough about Selenium to know what it *should* be called. On Sauce Labs they call it ``MicrosoftEdge`` (which is kind of different, it's not all lowercase and I'd have expected just ``edge`` since it's ``chrome`` not ``GoogleChrome``). Needle doesn't yet have support for it though so I've done:

```python
import bok_choy.browser

from needle.driver import NeedleWebDriverMixin
from selenium.webdriver.edge.webdriver import WebDriver as MicrosoftEdge

class NeedleMicrosoftEdge(NeedleWebDriverMixin, MicrosoftEdge):
    pass

bok_choy.browser.BROWSERS["MicrosoftEdge"] = NeedleMicrosoftEdge
```

benpatterson:
@dstufft Thanks for pointing that out. I see you have `selenium.webdriver.edge.webdriver` which fortunately for us is in Selenium now, which is great. 

Well, I'll merge this to at least save you a monkeypatch for Opera (that is, once we release this...which I can make happen faster if you need it).

@nedbat had sent a reference about a saucelabs tunnel...I'd like to chat about that if possible; I'm optimistic that's easy to add as well.

dstufft:
@benpatterson No rush, I'm using a forked bok_choy at the moment anyways with #146

Happy to discuss the SauceLabs tunnel too. Want that in another issue or IRC or?

PR #148, opened 2016-04-15
Make browser instantiation more resilient.

benpatterson:
The underlying selenium function that results in the "Can't load profile"
error is a 30-second polling function. Since that is also our default
timeout setting, bok-choy never gets a chance to retry. (That is, it takes
30 seconds for selenium to produce the WebDriverException). This gives
the driver another chance to create  browser instance when previously it
would only try once and fail. That was not the intention of the original
change with a `try_limit`. This also overrides the default timeout to
60 seconds due to the underlying timeout in selenium.

### DIFF
--- a/bok_choy/browser.py
+++ b/bok_choy/browser.py
@@ -207,7 +207,10 @@ def browser(tags=None, proxy=None):
             return False, None
 
     browser_instance = Promise(
-        browser_check_func, "Browser is instantiated successfully.", try_limit=3).fulfill()
+        # There are cases where selenium takes 30s to return with a failure, so in order to try 3
+        # times, we set a long timeout. If there is a hang on the first try, the timeout will
+        # be enforced.
+        browser_check_func, "Browser is instantiated successfully.", try_limit=3, timeout=95).fulfill()
 
     return browser_instance
 


benpatterson:
@jzoldak @raeeschachar @Muddasser 

This should help make the tests a bit more resilient against the Can't load profile error.

bok_choy/browser.py
```diff
@@ -207,7 +207,7 @@ def browser_check_func():
             return False, None
 
     browser_instance = Promise(
-        browser_check_func, "Browser is instantiated successfully.", try_limit=3).fulfill()
+        browser_check_func, "Browser is instantiated successfully.", try_limit=3, timeout=60).fulfill()
```
jzoldak: Ugh, yeah it looks like bok-choy is a bit clunky/confusing when you use try_limit, because it will stop [when the first of the limits (tries or time) is reached](https://github.com/edx/bok-choy/blob/master/bok_choy/promise.py#L113). :(

With an underlying 30-second polling function, with the timeout=60 I'm not sure that you will actually try 3 times either. I think it would be better to set this value even higher (120?) so that we are sure that the try_limit is the one that is expiring first. Together with a comment here that the try_limit, rather than the timeout value, is what we are expecting to kick in.

As an aside, I do think that this confusing bok-choy behavior is a bug. If you only specify try_limit I would assume that your intention is to actually use the value you gave it, not a mysterious default timeout. :(  We should discuss in an issue, because if we change the behavior now it could cause existing tests (not just ours, as edX is not the only org using bok-choy) to fail. Perhaps tho, we **should** change it before we get to v1.0 which hopefully will be soon.

benpatterson: @jzoldak Should we up this to 100 seconds to ensure it's tried thrice? That's pretty long, but we're at the hands of selenium's 30s timeout in at least one case.

jzoldak: 90 works for me.

I think I didn't really explain well what I was thinking in my prior comment. If I'm understanding correctly, as long as the value is > 3x30 for the timeout, it doesn't really matter what it is. The try_limit will be the one that kicks in anyhow.

benpatterson:
Merging based on discussions with @jzoldak

PR #149, opened 2016-04-15
multiprocess builds.

benpatterson:


jzoldak:
Closing because we've move from nose to pytest.

PR #151, opened 2016-04-21
Support python 3.5.1

benpatterson:


cgoldberg:
please update `setup.py` to include classifiers for py3:

```
'Programming Language :: Python :: 3',
'Programming Language :: Python :: 3.5',
```

doing so will indicate to users which versions of Python are supported, and will also trigger the python package index to mark the package as python 3 compatible, when released.

tests/test_performance.py
```diff
@@ -72,7 +72,7 @@ class ExplicitHarCaptureTest(HarCaptureTestBase):
     any environment setting. Note that this will only work if the `TestClass` is
     decorated, not the `test_case`.
     """
-    @expectedFailure
+    @skip  # Expected Failure. TODO.
```
cgoldberg: might be nice to annotate a "reason" for each skip, rather than using a bare skip.

such as:

@skip("expected failure")

cgoldberg:
@benpatterson ,
I don't quite understand the reason for these expected failure tests.  can you shed any light on why they exist at all?

benpatterson:
@cgoldberg We've used that pattern in other repos fwiw. But overall to your question, I'm not sure. Maybe @clytwynec knows?

cgoldberg:
@benpatterson my guess is they can be nuked, and whatever benefit they  provide can be done better with helper methods.

clytwynec:
These tests are a bit confusing.  The test itself is meant to fail, and the actual assertion is in the cleanup method (`self._inspect_har_content`).  Basically, by flagging it with `expectedfailure`, the test doesn't fail from the 'fake' failure in the test method.

cgoldberg:
>  by flagging it with expectedfailure, the test doesn't fail from the 'fake' failure in the test method.

so instead of creating a fake test that fails... why not just create a method that replicates the behavior of the failing test.  Then inside your test, swallow the exception that method raises, and inspect the har content.

benpatterson:
@clytwynec @cgoldberg I agree with your points, but I'm going to keep those out of scope for this change...just going to keep them there and not-run them. Admittedly, I am touching them and changing the nature of how they run...but I'm putting a cap on the work at that point so we can focus on the Py3 inclusion. 

I'll log a ticket in our backlog to clean them up later. Hopefully that sounds fair enough to you....

benpatterson:
@nedbat @dstufft @cgoldberg @jzoldak 

Looks like there are a couple of library upgrades needed, but this should put us into Py3 support. Any feedback on that? Note this is an extension of Donald's initial work.

clytwynec:
@benpatterson  Sounds good to me.

cgoldberg:
i see there is use of `abstractproperty` decorator in at least one place:
https://github.com/edx/bok-choy/blob/master/bok_choy/a11y/a11y_audit.py#L174

This will work fine, but `abstractproperty` has been deprecated since Python 3.3.  However, it's probably easier to leave the deprecated code in place than coming up with a workaround to support both 2 and 3.  (I don't think siz lib covers this case)

besides that, LGTM.

jzoldak:
Implemented in #168

PR #157, opened 2016-06-10
Version bump to include a11y-rules in platform (v0.5.3)

clrux:
@benpatterson I guess I should've included this work with the previously merged PR. Doh! The working example I tested with was using a manual Git commit hash/egg, but in order to make this useful to other developers it needs to be permanent :)

- [ ] @benpatterson

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,3 +14,4 @@ Jonathan Piacenti <kelketek@gmail.com>
 Christina Roberts <christina@edx.org>
 Bill Agee <billagee@gmail.com>
 Carol Tong <ctong@edx.org>
+Chris Rodriguez <edx@chrisrodriguez.me>
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+v0.5.3 (06/07/16)
+* Updated the a11y-custom-rules output to include the severity of the a11y failure
+* Added Chris Rodriguez to AUTHORS file
+
 v0.5.2 (04/25/16)
 * Support Opera browser
 * Improve browser instantiation logic to include retries
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -38,9 +38,9 @@ copyright = u'2016, EdX'
 # built documents.
 #
 # The short X.Y version.
-version = '0.5.2'
+version = '0.5.3'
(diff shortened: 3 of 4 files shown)


jzoldak:
We'd also need to bump the version in docs/conf.py.
I'm not sure why that test_har_is_captured_on_success_in_auto_mode test failed on travis. :|

clrux:
@jzoldak Do you want me to look into this failure? Ben says you may be considering the Har stuff.

jzoldak:
@benpatterson @clrux 
I've confirmed by [retriggering a build of master on travis](https://travis-ci.org/edx/bok-choy/builds/135838363) that master currently has the same failure.
We have a few options:
* ignore the failure and merge anyway. master will remain broken. :(
* skip the test "for now" until we get a chance to fix it. I did a bit of digging and it looks like that test is actually the only one that ever looks at the har file content. See [inspect_har_content](https://github.com/edx/bok-choy/blob/master/tests/test_performance.py#L55). Unfo I don't have a development environment that has bmp installed correctly and it would take a bit of time troubleshooting to dig into the exact difference.
* wait until next week to decide what to do and it might be a no-op. I'm expecting to show bok-choy some love in the upcoming hackathon later this week, in which I was going to rip out the HAR stuff anyway.

clrux:
I'd love to merge this despite the error. It's unrelated and it'd unblock a few things on the platform end.

benpatterson:
I spent a cycle looking at the har capture. Indeed it is broken. We capture hars, which is great, but we don't capture the waterfalls, so the index out of range error is appropriate. As best I can tell, some things were changed in firefox with [version 41](https://developer.mozilla.org/en-US/docs/Tools/Network_Monitor#CopySave_All_As_HAR). Possibly we need to use the [HAR export trigger](http://www.softwareishard.com/blog/har-export-trigger/) to get the full data. Though, I'm still scratching my head on this a bit, since we indeed are capturing HAR files.

Either way, I think this is blocking progress for @clrux and I think we should merge even with the failure, or skip the test. I'm going to finish off this request momentarily.

benpatterson:
@clrux the new version of bok-choy has been uploaded to pypi

clrux:
Thanks!

PR #159, opened 2016-06-16
Remove browsermobproxy and HAR capture feature

jzoldak:
@thallada @ajpal @jmbowman @dianakhuang


test_requirements.txt
```diff
@@ -2,4 +2,5 @@ coverage==3.7
 nose==1.3.0
 mock==1.0.1
 pep8==1.7.0
+requests==2.10.0
```
dianakhuang: Why is this being added?

dianakhuang:
Other than that one question, 🚢 it!

test_requirements.txt
jzoldak: Previously it got installed "for free" as a subdependency of browsermob-proxy, whose setup.py includes this:
```
install_requires=['requests>=1.1.0'],
```

jmbowman:
👍

test_requirements.txt
jzoldak: ... and then @jmbowman discovered that it's not even being used in the test file that imported it so we could get rid of it anyhow. Bonus!

benpatterson:
🎆

PR #160, opened 2016-06-16
Upgrade edx_lint

ajpal:
@jzoldak @dianakhuang @jmbowman @thallada

### DIFF
--- a/bok_choy/javascript.py
+++ b/bok_choy/javascript.py
@@ -87,6 +87,7 @@ def _decorator(store_name, store_values):
 
         # Add a `wait_for_js` method to the class
         if not hasattr(clz, 'wait_for_js'):
+            # pylint: disable= literal-used-as-attribute
             setattr(clz, 'wait_for_js', _wait_for_js)
 
         # Store the RequireJS module names in the class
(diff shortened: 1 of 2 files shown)


bok_choy/javascript.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable= literal-used-as-attribute
```
dianakhuang: Instead of disabling this for the whole file, I think it's better to disable this on the line where the pylint error was showing up, so we know that we are only doing it for those specific cases.

ajpal: @dianakhuang fixed

dianakhuang:
🚀 🚢 📦

bok_choy/javascript.py
```diff
@@ -87,6 +87,7 @@ def decorator(clz):  # pylint: disable=missing-docstring
 
         # Add a `wait_for_js` method to the class
         if not hasattr(clz, 'wait_for_js'):
+            # pylint: disable= literal-used-as-attribute
```
jzoldak: nit: Remove the space after the =
nit: In this case it doesn't matter because we're disabling from the last statement in the branch. Generally though I'd recommend putting the pragma on the same line (with 2 spaces before the # symbol, otherwise pep8 will complain) which will disable for only that line.

PR #161, opened 2016-06-16
Add FIREFOX_PROFILE_PATH option to override default profile

thallada:
Allows overriding the default firefox profile that bok-choy creates with one created by the user. This is useful for setting profile preferences needed for a specific set of tests (e.g. disabling the firefox warning pop-up for a script taking too long for very long running javascript tests).

@ajpal @dianakhuang @jmbowman @jzoldak

### DIFF
--- a/bok_choy/browser.py
+++ b/bok_choy/browser.py
@@ -47,6 +47,8 @@ BROWSERS = {
     'opera': NeedleOpera,
 }
 
+FIREFOX_PROFILE_ENV_VAR = 'FIREFOX_PROFILE_PATH'
+
 
 class BrowserConfigError(Exception):
 
@@ -159,13 +161,14 @@ def browser(tags=None, proxy=None):
     1. Local browsers: If the proper environment variables are not all set for the second case,
         then we use a local browser.
 
-        * The environment variable `SELENIUM_BROWSER` can be set to
-          specify which local browser to use. The default is Firefox.
-        * Additionally, if a proxy
-          instance is passed and the browser choice is either Chrome or Firefox, then the browser will
-          be initialized with the proxy server set.
-        * The environment variable `SELENIUM_FIREFOX_PATH` can be used for specifying a path to the
-          Firefox binary. Default behavior is to use the system location.
+        * The environment variable `SELENIUM_BROWSER` can be set to specify which local browser to use. The default is \
+          Firefox.
+        * Additionally, if a proxy instance is passed and the browser choice is either Chrome or Firefox, then the \
+          browser will be initialized with the proxy server set.
+        * The environment variable `SELENIUM_FIREFOX_PATH` can be used for specifying a path to the Firefox binary. \
+          Default behavior is to use the system location.
+        * The environment variable `FIREFOX_PROFILE_PATH` can be used for specifying a path to the Firefox profile. \
+          Default behavior is to use a barebones default profile with a few useful preferences set.
 
     2. Remote browser (not SauceLabs): Set all of the following environment variables, but not all of
         the ones needed for SauceLabs:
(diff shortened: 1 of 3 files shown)


benpatterson:
Hi @thallada. I touched browsers.py recently for the firefox path changes. The one you propose here could be very useful! So I jumped in to take a look...

Could you add tests please?

[Here](https://github.com/edx/bok-choy/blob/master/tests/test_browser.py#L53-L56) is an example with the firefox binary path that could help you get started.

thallada:
@benpatterson added a couple of tests for this change. Thanks for pointing out that example for me!

bok_choy/browser.py
```diff
@@ -159,13 +161,14 @@ def browser(tags=None, proxy=None):
...
         then we use a local browser.
 
-        * The environment variable `SELENIUM_BROWSER` can be set to
-        specify which local browser to use. The default is Firefox.
-        * Additionally, if a proxy
-        instance is passed and the browser choice is either Chrome or Firefox, then the browser will
-        be initialized with the proxy server set.
-        * The environment variable `SELENIUM_FIREFOX_PATH` can be used for specifying a path to the
-        Firefox binary. Default behavior is to use the system location.
+        * The environment variable `SELENIUM_BROWSER` can be set to specify which local browser to use. The default is \
```
jzoldak: We are within a docstring here so there's no need to put continuation markers at the ends of the lines.

bok_choy/browser.py
```diff
@@ -260,17 +263,24 @@ def _local_browser_class(browser_name):
...
-            firefox_profile.set_preference('media.navigator.permission.disabled', True)
-
-            # Disable the initial url fetch to 'learn more' from mozilla (so you don't have to
-            # be online to run bok-choy on firefox)
-            firefox_profile.set_preference('browser.startup.homepage', 'about:blank')
-            firefox_profile.set_preference('startup.homepage_welcome_url', 'about:blank')
-            firefox_profile.set_preference('startup.homepage_welcome_url.additional', 'about:blank')
+            if profile_dir:
+                LOGGER.info("Using firefox profile: %s", profile_dir)
+                firefox_profile = webdriver.FirefoxProfile(profile_dir)
```
jzoldak: What will happen if the profile directory does not exist, is not readable, or its contents are not in a valid format?

thallada: If the directory does not exist or is not readable, selenium will throw an OSError exception that makes it fairly obvious that the profile directory is bad. If it's empty, that's okay, selenium creates an empty directory for the default profile. I tried using a profile with a bunch of garbage files, but firefox seems to tolerate that okay and still work.

bok_choy/browser.py
thallada: I added those because the rendered HTML was treating the newlines in the docstring as the end of the list element and that was breaking the list formatting (see http://bok-choy.readthedocs.io/en/latest/api_reference.html#bok_choy.browser.browser)

bok_choy/browser.py
jzoldak: Cool. Let's wrap this in a try/catch block that captures the OSError exception and raises a BrowserConfigError which tells the person explicitly that the directory does not exist or is not readable.

benpatterson: FWIW I don't think a try/catch block is necessary. We don't use one for the SELENIUM_FIREFOX_PATH setting because the naturally-raised exception is clear enough that the file doesn't exist. 

I see @thallada has already made this change, but he doesn't have to undo it if you feel strongly about raising a specific exception.

bok_choy/browser.py
jzoldak: OIC. Nice, I didn't realize that happened. Good catch!

jzoldak:
👍 🚀

jmbowman:
Could we do this in a way that provides programmatic access to the profile object?  While it's sometimes useful to set a different profile directory, my primary use case for this functionality would be to call [set_preference()](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_firefox/selenium.webdriver.firefox.firefox_profile.html#selenium.webdriver.firefox.firefox_profile.FirefoxProfile.set_preference) on the FirefoxProfile object in order to customize the settings [like this](http://stackoverflow.com/a/19151816).

thallada:
We can access the profile in a test with `browser.profile.set_preference()`, we'd need a separate change to allow using the same set of custom preferences over multiple test cases. Maybe some sort of self.profile property that would be part of the WebAppTest object that would be passed into the browser creation in setUp.

PR #165, opened 2016-06-16
Adds save_source call to _save_artifacts procedure

### LINKED ISSUE #86: Add html source file of failed test along with screenshot in reports
Right now we are adding the screenshot of bokchoy failed tests in reports. I guess it would be beneficial to the developer if we add the html source to a file of failed test. We should take the html source of page when we are taking the screenshot.

thallada:
This should close #86. Will save the HTML source of the page along with the screenshot if the test fails or errors.

@ajpal @jzoldak @jmbowman @dianakhuang

### DIFF
--- a/bok_choy/web_app_test.py
+++ b/bok_choy/web_app_test.py
@@ -8,7 +8,7 @@ from uuid import uuid4
 
 from needle.cases import NeedleTestCase, import_from_string
 
-from .browser import browser, save_screenshot, save_driver_logs
+from .browser import browser, save_screenshot, save_driver_logs, save_source
 
 
 class WebAppTest(NeedleTestCase):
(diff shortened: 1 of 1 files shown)


jzoldak:
@thallada couple small things:
* update the docstring for the _save_artifacts method
* we need to figure out why coveralls didn't report back. From the [travis build](https://travis-ci.org/edx/bok-choy/builds/138198329), scroll down to the end and open the lines beneath "$ coveralls". It's saying `Failure to gather coverage: Couldn't read data from '/home/travis/build/edx/bok-choy/.coverage': CoverageException: Doesn't seem to be a coverage.py data file
'url'`

jzoldak:
👍 🚀

PR #166, opened 2016-06-16
Use pytest to run tests

jmbowman:
Use pytest by default to run tests, and update the dependencies accordingly.  This also eliminates the need for the run_tests.sh.


jzoldak:
👍 😍

PR #170, opened 2016-06-17
v0.6.0

jzoldak:
@benpatterson 

New version. The next one will be 1.0!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,10 @@
+v0.6.0 (06/17/16)
+* Support Python 3
+* Use pytest to run tests
+* Save browser page source on failure
+* Allow configuration of the firefox profile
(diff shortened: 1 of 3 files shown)


CHANGELOG
```diff
@@ -1,3 +1,10 @@
+v0.6.0 (06/17/16)
+* Support Python 3
+* Use pytest to run tests
+* Save browser page source on failure
+* Allow configuration of the firefox profile
```
benpatterson: nit: I'd reword it to something like "allow firefox profile to be set from a file on disk" or something like that, that makes it clear it's an already-saved profile, rather than one which is programmatically created.

benpatterson:
oh, ah, i didn't realize it was an FYI. But...6.0 woohoo!!!

PR #175, opened 2016-07-25
Support running tests in parallel

jmbowman:
Added support for running tests in parallel via pytest-xdist.  On my local machine, this impacts test duration with increasing number of worker processes as follows:

1 node: 361.16 seconds
2 nodes: 216.78 seconds
3 nodes: 169.27 seconds
4 nodes: 147.60 seconds
5 nodes: 145.69 seconds
6 nodes: 148.79 seconds

Travis VMs oficially have 1.5 cores, so this should see a little bit of a speedup there also.  Note that a separate HTTP server is launched for each test process; trying to instantiate just one to share between them didn't seem to fit well with the startup hook sequence (partially because pytest-xdist is also intended to support distributing work across test processes on different machines).

Other changes made:

* List the 10 slowest test cases from each run so optimization efforts can be focused on them
* Updated a few dependencies for test runs
* Updated the README to reflect that the latest Firefox and selenium currently work together

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ install:
 - pip install -r requirements/tox.txt
 - pip install tox-travis
 script:
-- tox
+- tox -- -n 2 --durations=10
 - tox -e doc
 after_success:
 - pip install coveralls
--- a/README.rst
+++ b/README.rst
@@ -30,14 +30,16 @@ Running Tests
 
 To run the test suite for bok-choy itself:
 
-* Install Firefox, `version 46.0.1 <https://ftp.mozilla.org/pub/firefox/releases/46.0.1/>`_
-  or earlier (later versions require a different selenium driver)
+* Install Firefox; as of this writing, the current `version 47.0.1 <https://ftp.mozilla.org/pub/firefox/releases/47.0.1/>`_
+  works with the latest selenium Python package (2.53.6)
 * Install `phantomjs <http://phantomjs.org/download.html>`_
 * Create a virtualenv which uses Python 2.7 (or Python 3.5)
 * With that virtualenv activated, run ``pip install -r requirements/tox.txt`` to
   install the `tox <http://tox.testrun.org/>`_ testing tool and its
   dependencies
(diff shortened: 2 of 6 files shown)


jmbowman:
@benpatterson Here's the PR for parallel bok-choy testing that I mentioned this morning.

.travis.yml
```diff
@@ -16,7 +16,7 @@ install:
 - pip install -r requirements/tox.txt
 - pip install tox-travis
 script:
-- tox
+- tox -- -n 2 --durations=10
```
benpatterson: What does `durations` set? I couldn't find that documented.

jmbowman: It's a pytest option for listing the N slowest tests from the run: http://doc.pytest.org/en/latest/usage.html#profiling-test-execution-duration

jmbowman: Sample output:

```
=========================== slowest 10 test durations ============================
33.28s call     tests/test_javascript.py::JavaScriptTest::test_wait_for_defined_failure
15.52s call     tests/test_helpers.py::WaitHelpersTest::test_element_presence_wait
15.47s call     tests/test_helpers.py::WaitHelpersTest::test_element_visibility_wait
10.55s call     tests/test_helpers.py::WaitHelpersTest::test_element_invisibility_wait
10.34s call     tests/test_helpers.py::WaitHelpersTest::test_element_absence_wait
9.35s call     tests/test_delay.py::DelayTest::test_delay
8.45s call     tests/test_javascript.py::JavaScriptTest::test_wait_for_defined_with_reload
5.73s call     tests/test_javascript.py::JavaScriptTest::test_wait_for_defined
5.30s call     tests/test_ajax.py::AjaxTest::test_ajax
5.13s call     tests/test_delay.py::SlowTest::test_slow
```

benpatterson: Ah thank you for explaining. Nose has a timeout concept, which is what I originally thought this was.

jzoldak:
👍

jzoldak:
@benpatterson are we good to merge this in?

benpatterson:
:+1:

PR #177, opened 2016-09-30
Add firefox logging

jzoldak:
@edx/testeng 

This makes debugging issues with firefox browser startup easier. 
Especially useful given the trickiness of finding firefox and selenium versions that work together.


jzoldak:
Turns out that in practice, you need to pass a file object, not a string with the pathname.
I don't have time to code that up nicely (including hardening for the directory doesn't exist, you don't have permissions, thinking of how to clean up the file or if you really want an empty file every time even when there are no errors, etc.), so I'm closing this for now.

PR #178, opened 2016-11-01
Make all Bok Choy tests show long failure messages

andy-armstrong:
@jzoldak @benpatterson this is a simple change to improve failure messages for Bok Choy failures. See this Stack Overflow article for the reasoning:

http://stackoverflow.com/questions/4634625/python-can-unittest-display-expected-and-actual-values

This has been bothering me for a while, when we see Jenkins failures with no indication of what the actual and expected values were.


andy-armstrong:
From discussions with edX developers, I'm closing this as the approach is too rigid. I'll investigate introducing a new base class for edX acceptance tests in edx-platform where this can live.

PR #179, opened 2016-12-02
Update Sphinx docs to edX theme, fix doc test

bmedx:
PLAT-1087 - Update the Sphinx docs to use the theme from the  edx-sphinx-theme repo. Also ran into an issue with the docs tests that is unrelated but may have failed on commit so I fixed that too. Problem just seems to have been that GitHub changed their html layout.

@jmbowman @jzoldak - please take a look when you get a chance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,9 @@ nosetests.xml
 .project
 .pydevproject
 
+# PyCharm
+.idea
+
 # Sphinx
 docs/_build
 docs/**/*.png
--- a/docs/code/round_3/pages.py
+++ b/docs/code/round_3/pages.py
@@ -21,7 +21,7 @@ class GitHubSearchResultsPage(PageObject):
         """
         Return a list of results returned from a search
         """
-        return self.q(css='ul.repo-list > li > h3.repo-list-name > a').text
+        return self.q(css='ul.repo-list > li > div > h3 > a').text
 
 
 class GitHubSearchPage(PageObject):
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 
-import sys, os
+import os
+import sys
+import edx_theme
+
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -35,6 +35,9 @@ nosetests.xml
 .project
 .pydevproject
 
+# PyCharm
+.idea
```
benpatterson: nit: should this be `.idea/` ?

benpatterson:
Assuming you don't discover anything, 👍 from me

.gitignore
bmedx: Thanks, I will run checks on Read the Docs today, I didn't know we could run it on branches. For `.idea` my understanding is that without the trailing slash it will match files, directories, symlinks, etc. by that name. I figured we didn't want any of those things if they should exist, but if there's a reason why we might want to let one of those through I'd be happy to change it.

bmedx:
@jmbowman @benpatterson I had to make an adjustment to setup.py to avoid some version problems with selenium that may be impacting people trying to install that way (Read the Docs uses setup.py, which is how this relates). I verified the issue does not impact pip installs. Still, pegging versions in setup.py isn't best practice so I wanted to call this out for discussion if you have any concerns.

jmbowman:
I'd rather go with the same constraint bounds as needle than specify an exact version.  Also, a fix for selenium 3 support has been checked into the master branch of needle if we'd rather just specify a commit hash in the requirements file until the next release comes out.  And don't we want to [configure RTD to use a requirements file](https://github.com/edx/edx-sphinx-theme#read-the-docs-configuration) anyway?

bmedx:
We are configured to use RtD with requirements. The problem is that it hits setup.py first, installs selenium 3, then fails on needle before it even gets there. Bok-choy requirements is using the exact version for selenium so I believe it would just get slammed to 2.53.6 anyway, if it got that far.

The differences between selenium 2 and 3 are significant, so I don't want to open that can of worms for this particular task. I'll open the bounds to be <3, >2 in case they do another 2.x release and comment requirements to make it clear that we'll need to change things in both places if we go to 3+.

benpatterson:
CC @jzoldak 

Yeah, selenium 3 is something we need to support sooner rather than later. The needle tie-in becomes problematic in that view. Needle doesn't exactly look like it's getting a lot of support right now. We primarily use needle for screenshot assertions, which have proven to be brittle, but--that said--accurate and could be quite useful if used surgically.

bmedx:
@jmbowman do the current changes seem sufficient? Everything built and is running correctly on RtD.

jmbowman:
Certainly good enough for a doc update and build fix.  We can sort out the needle and selenium 3 stuff later with a dedicated ticket and PR for that.

PR #180, opened 2016-12-19
v0.6.1

jzoldak:
@edx/testeng @Muddasser

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+v0.6.1 (12/19/16)
+* Add support for programmatic customization of the Firefox profile preferences
+* edX best practices - Update doc theme; add yaml file for OEP2 compliance tracking
+* Improve linting and test execution
+
 v0.6.0 (06/17/16)
 * Support Python 3
 * Use pytest to run tests
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,9 +42,9 @@ author = edx_theme.AUTHOR
 # built documents.
 #
 # The short X.Y version.
-version = '0.6.0'
+version = '0.6.1'
 # The full version, including alpha/beta/rc tags.
-release = '0.6.0'
+release = '0.6.1'
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 import codecs
 from setuptools import setup
 
-VERSION = '0.6.0'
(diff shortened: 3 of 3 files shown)


jzoldak:
@jmbowman I think that the python 3.5 build might be having a race condition issue with Needle creating directories when running tests in multiprocess mode?

jmbowman:
Looks like it.  Python 3.2 added an optional `exist_ok` parameter to `makedirs` to handle cases like this, but it isn't available in 2.7 and would require a `needle` code change anyway.  It looks like it tries to grab the directory paths from environment variables and picks deterministic defaults, maybe we should just create those directories preemptively in `tox.ini`?

jmbowman:
We should also create a `needle` issue or PR to fix the bug; it hasn't had a new release in a while, but there have been recent commits to the repo.

cgoldberg:
oh the irony.  it looks like needle is releasing 0.4 *today*, but the race condition remains.

cgoldberg:
I just looked at the needle source code, and it's def not safe to run from multiple processes (for any version of python).

For the race condition, see: https://github.com/bfirsh/needle/blob/1daace939df6829febe663c630e9e70ec20b779d/needle/cases.py#L112

If anybody wants to make a PR to needle, here is what needs to be done...

The code in question:

```
for i in (self.baseline_directory, self.output_directory):
    if not os.path.exists(i):
        print('Creating %s' % i, file=sys.stderr)
        os.makedirs(i)
```

so, it's attempting to guard against creating duplicate directories by first checking `os.path.exists`.  The race condition happens when that check comes back as false and then before it has a chance to create the directory, another process calls `os.makedirs` to create a directory with the same name.  This is why LBYL is usually inferior to EAFP style in Python :)

That snippet from needle could be re-written in EAFP style so it's not affected by a race condition:

```
for dirname in (self.baseline_directory, self.output_directory):
    try:
    	print('Creating {}.format(dirname)' file=sys.stderr)
    	os.makedirs(i)
    except OSError:
    	pass
```

note: the try/catch has the same effect as the `exist_ok` parameter that was added to `makedirs` in Python 3.2

jzoldak:
Oh, nice. @jmbowman @cgoldberg thx for narrowing that down!
I went quick and dirty and just merged #181 to not run tests in parallel for now to get this new version released.
In a separate thread ;) - I'll submit a PR upstream to Needle.

jzoldak:
Reference: Upstream [PR #57](../../../bfirsh/needle/pull/57) for the race condition.

PR #181, opened 2016-12-28
Do not run tests in parallel because Needle is not thread safe

jzoldak:
and thus bok choy is neither. :(

See the discussion on #180


benpatterson:
👍

PR #183, opened 2017-01-10
Firefox improvements in logging and default settings

jzoldak:
* Add firefox logging
* Firefox default settings improvements in speed and network usage
  * Disable fetching an updated version of firefox
    **Note**: by default this occurs even though we never will  choose to install the new version
  * Disable plugin checking
  * Disable health reporter
  * Disable all data upload (Telemetry and FHR)
  * Disable crash reporter

@edx/testeng

### DIFF
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -40,6 +40,11 @@ class TestBrowser(TestCase):
         self.assertEqual(preferences['browser.startup.homepage'], 'about:blank')
         self.assertEqual(preferences['startup.homepage_welcome_url'], 'about:blank')
         self.assertEqual(preferences['startup.homepage_welcome_url.additional'], 'about:blank')
+        self.assertFalse(preferences['app.update.enabled'])
+        self.assertTrue(preferences['plugins.hide_infobar_for_outdated_plugin'])
+        self.assertFalse(preferences['datareporting.healthreport.service.enabled'])
+        self.assertFalse(preferences['datareporting.policy.dataSubmissionEnabled'])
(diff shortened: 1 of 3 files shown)


tests/test_browser.py
```diff
@@ -81,6 +86,21 @@ def test_no_custom_firefox_path(self):
...
+
+    @patch.dict(os.environ, [('SELENIUM_FIREFOX_LOG', '')])
+    def test_no_custom_firefox_log(self):
+        browser_kwargs_tuple = bok_choy.browser._local_browser_class('firefox')  # pylint: disable=protected-access
+        assert 'firefox_binary' not in browser_kwargs_tuple[2]
+
+    @patch.dict(os.environ, [('SELENIUM_FIREFOX_PATH', '/foo/path'), ('SELENIUM_FIREFOX_LOG', '/foo/file.log')])
+    def test_custom_firefox_path_and_log(self):
+        browser_kwargs_tuple = bok_choy.browser._local_browser_class('firefox')  # pylint: disable=protected-access
+        assert 'firefox_binary' in browser_kwargs_tuple[2]
```
jmbowman: Looks like there are now 4 different tests with the same content; break out into a utility function?  Should there be further assertions on the FirefoxBinary object attributes to make sure it's configured correctly for each scenario?

jzoldak:
@jmbowman Refactored the tests. We're using private methods for the verifications, so there is some chance that a selenium upgrade will break the tests. I feel generally OK about this, especially as we were already using _local_browser_class('firefox').

PR #185, opened 2017-03-23
Miscellaneous cleanup

jmbowman:
An assortment of small fixes I made while working on some dependency upgrades, grouped together here to avoid distracting from the main content of forthcoming PRs:

* Fix a test broken by a change in GitHub's search page markup
* Have git ignore more artifacts generated by tests
* Don't try to include a non-existent requirements file in the built package
* Update some copyright years
* Fix some doc generation warnings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,5 +46,7 @@ docs/**/*.png
 .cache
 screenshots
 logs
+geckodriver.log
 ghostdriver.log
 server.log
+tests.*.png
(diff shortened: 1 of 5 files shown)


jmbowman:
I originally tried to lump in the addition of Python 3.6 testing because it looked small, but that blew up because pylint doesn't officially support it yet (although it sounds like they'll be putting out a release that does anytime now).  So this is now just some assorted housecleaning before I start creating PRs for selenium and Firefox upgrades.

PR #186, opened 2017-03-27
Support selenium 3.x and newer Firefox versions

jmbowman:
Made a few changes to support newer selenium and Firefox versions without removing support for older ones:

* Upgrade needle, which has recently had a number of fixes committed for this.  A new version is expected in the next few days, but this git commit from master lets us move forward with testing in the meantime.  (We can wait until the official needle release before cutting the next bok-choy release.)
* Relax the upper version constraint on selenium in setup.py.
* Work around a PhantomJS shutdown bug which hasn't yet been fixed in the selenium package.
* Update the Travis configuration to use Firefox 52.0.1 and geckodriver.
* Add some time for xvfb to finish starting up in Travis runs (the [Travis documentation](https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI) recommends this, and needle's test suite needed it).

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -3,11 +3,7 @@
 # needle (selenium is also used directly)
 nose==1.3.7
 Pillow==3.3.0
-
-# Note that to get setup.py to work correctly with needle we have
-# version constraints specified in setup.py as well. We will need
-# to change both to upgrade to selenium 3 or above.
-selenium==2.53.6
+selenium==3.3.1
 
 # Then the direct dependencies
 
@@ -15,7 +11,7 @@ selenium==2.53.6
 lazy==1.2
 
 # For test assertions involving screenshots and visual diffs
-needle==0.4.1
+needle==0.5.0
 
 # For Python 2 & 3 compatibility
 six==1.10.0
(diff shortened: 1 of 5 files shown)


requirements/base.txt
```diff
@@ -3,19 +3,15 @@
...
+selenium==3.3.1
 
 # Then the direct dependencies
 
 # For lazy evaluation of the page object's a11y_audit attribute
 lazy==1.2
 
 # For test assertions involving screenshots and visual diffs
-needle==0.4.1
+git+https://github.com/bfirsh/needle.git@2bc898654c5dc887d47a3ba26be92a6d42113440#egg=needle
```
jzoldak: I hesitate to do this as we are not under control of the repo, and this commit could disappear. Acknowledging that the risk is pretty low, still I recommend that we either wait or fork the repo (either under your account if it's easier since it's temporary or the edx org).

jmbowman: The official needle release should be out [any time now](https://github.com/bfirsh/needle/pull/70#issuecomment-289307058), so I'm ok with holding off on merging this until that's available.  I just needed something checked into a branch so I could refer to it in requirements files for local testing (which is currently in progress), and didn't want to hold up code review for it.

jmbowman:
@jzoldak needle 0.5.0 has been released, updated the PR accordingly.

PR #187, opened 2017-04-03
Catch WebDriverExceptions while waiting for PageObject conditions

jmbowman:
Catch WebDriverExceptions during `PageObject.wait_for()` and all derived methods, to avoid failing tests due to side effects of expected page refreshes.  This should help resolve problems like [PLAT-1200](https://openedx.atlassian.net/browse/PLAT-1200).

This is about the minimum change required to accomplish this; we could refactor it a little more to group together and rename `no_error` and `no_selenium_error` if you think it's appropriate.  I just think we should at least leave an import of the `no_error` function in its current module and name for backwards compatibility (since there's nothing to indicate it as private).

I've already tested edx-platform with these changes, and only got a few failures which seem unrelated (they're somewhat flaky even with the bok-choy version currently in use).  I'm digging into those further to see if I can fix them completely.

### DIFF
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -24,6 +24,8 @@ from .promise import Promise, EmptyPromise, BrokenPromise
 from .a11y import AxeCoreAudit, AxsAudit
 
 
+LOGGER = logging.getLogger(__name__)
+
 # String that can be used to test for XSS vulnerabilities.
 # Taken from https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#XSS_Locator.
 XSS_INJECTION = "'';!--\"<XSS>=&{()}"
@@ -79,6 +81,7 @@ def no_selenium_errors(func):
         try:
             return_val = func(*args, **kwargs)
         except WebDriverException:
+            LOGGER.warning(u'Exception ignored during retry loop:', exc_info=True)
             return False
         else:
             return return_val
--- a/bok_choy/query.py
+++ b/bok_choy/query.py
@@ -3,6 +3,8 @@ Tools for interacting with the DOM inside a browser.
 """
 from __future__ import absolute_import
 
+import logging
+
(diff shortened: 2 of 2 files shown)


jmbowman:
@edx/testeng Anyone up for reviewing this?

benpatterson:
Basic question, but can you explain what would happen on a real failure?

jmbowman:
It basically treats any WebDriverException (or subclass thereof) as a failure of the wait condition, and keeps retrying until the timeout.  If the timeout is reached, the test fails with the error message that was declared for the wait (for example, "Promise not satisfied: Waiting for new post button").  It currently doesn't log the individual swallowed exceptions, since they often don't indicate actual problems.  Any output from the browser or driver related to the error would be logged, though.

benpatterson:
I'm still thinking through this. I prefer to see the errors, but then again
having a failed wait_for condition might be more straightforward when it
comes to troubleshooting. @jzoldak @estute what do you think? I'm kind of
leaning towards a thumbs-up on this approach but need to think it through.

-- 
*Ben Patterson*

edX | Engineering Manager  | ben@edx.org

141 Portland Street, 9th floor

Cambridge, MA 02139
http://www.edx.org <http://www.edxonline.org/>

[image:
http://www.e-learn.nl/media/blogs/e-learn/edX_Logo_Col_RGB_FINAL.jpg?mtime=1336074566]

jmbowman:
After discussion at standup this morning, I adjusted this to log the ignored exceptions as warnings instead of swallowing them outright; that preserves the information but clearly indicates that the exceptions probably weren't the immediate cause of test failure.  Does this work better for everyone?  I'd like to get this merged and a release issued so we can start getting the corresponding test reliability improvements in edx-platform.

benpatterson:
I'm fine with it but would like to hear @jzoldak 's opinion. Thanks for pinging on it.

PR #190, opened 2017-04-13
Teach bokchoy how to tack new queries onto existing ones

### LINKED ISSUE #45: Add Support for Query Chaining
After retrieving a parent element I would like to be able to make queries on the child elements. jQuery and Selenium offer real-world implementation examples.

``` html
<div class="parent">
  <div class="child"></div>
  <div class="child"></div>
</div>
```

``` python
page = ...

# Get the parent element
parent = page.q(css='div.parent')

# Get all of the children nested in the parent element
children = parent.q(css='div.child')

# Get the first child element
child = children[0]

# Infinite chaining (yes, the call below should fail since there is no span)
child.q(css='span')
```

cpennington:
Fixes #45

### DIFF
--- a/bok_choy/query.py
+++ b/bok_choy/query.py
@@ -10,6 +10,7 @@ from collections import Sequence
 from itertools import islice
 from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.common.by import By
+from selenium.webdriver import ActionChains
 import six
 from bok_choy.promise import Promise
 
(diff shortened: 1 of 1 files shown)


feanil:
Closing stale old PR.

openedx-webhooks:
Although this pull request is already closed,  I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #192, opened 2017-06-15
Disable new Firefox JSON Viewer

jmbowman:
Starting in Firefox 53, raw JSON responses are now displayed in a JSON viewer widget by default.  This is convenient for development, but breaks automated tests which have historically expected JSON to be displayed in the browser as raw text (like many of the tests in edx-platform).  I'm hard pressed to see this viewer being useful in automated tests, so this change disables it by default for new Firefox sessions.


jmbowman:
@edx/testeng This is ready for review.

PR #196, opened 2017-09-07
Add details about tox/travis testing to docs.

doctoryes:
@jmbowman How about something like this?

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -1,0 +1,36 @@
+Testing Environment Configuration
+=================================
+
+Testing via TravisCI
+--------------------
+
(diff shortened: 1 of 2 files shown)


docs/testing.rst
```diff
@@ -0,0 +1,36 @@
+Testing Details
```
jmbowman: Maybe something like "Test Environment Configuration" would be more accurate?  Otherwise, looks great.

doctoryes:
@jmbowman Any ideas about the test failure in Python 2.7? Is it something flaky that's occurring?

jmbowman:
You definitely seem to be hitting some flakiness; the previous commit had 2 different failures under 3.5 but 2.7 passed.  More justification for tackling [PLAT-1713](https://openedx.atlassian.net/browse/PLAT-1713) soon.

PR #199, opened 2018-01-25
update selenium and FF

estute:
In order to run bokchoy tests with FireFox 59.0.1, which contains both the reworked quantum engine and meltdown fixes, selenium 3.11.0 is required. This also upgrades the versions of Firefox and geckodriver used in the travis file


jmbowman:
@estute It looks like equivalent code has now been merged to master; can this be safely closed?

PR #200, opened 2018-03-01
WL-1435 Modified wait for page to handle intermittent page load failures

kashifch:
Recently we have seen a lot of e2e and microsites test failures with the following message

BrokenPromise: Promise not satisfied: The document and all sub-resources have finished loading.

Upon investigating we found out that on stage the page is loaded but somehow the third party requests(there are a lot of requests sent to optimizely, google analytics etc.) get stuck and since the wait for page relies on the document.readyState to become "complete" it throws and error and the test is considered as failure.

I have used an experimental approach where we added an additional check if document does not reach the ready state, instead of throwing error we check if document.readyState="interactive", if so we continue with the test instead of throwing error.

This approach can take care of these errors where page is loaded but error is raised due to some third party call getting stuck.

Here is a reference to the ticket with details about it 
[WL-1435](https://openedx.atlassian.net/browse/WL-1435)

### DIFF
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -423,6 +423,14 @@ class PageObject(object):
         Raises:
             BrokenPromise: The timeout is exceeded without the page loading successfully.
         """
+
+        def _is_document_interactive():
+            """
+            Check the loading state of the document to ensure the document is in interactive mode
+            """
+            return self.browser.execute_script(
+                "return document.readyState=='interactive'")
+
         def _is_document_ready():
             """
             Check the loading state of the document to ensure the document and all sub-resources
@@ -431,11 +439,25 @@ class PageObject(object):
             return self.browser.execute_script(
                 "return document.readyState=='complete'")
 
-        EmptyPromise(
-            _is_document_ready,
-            "The document and all sub-resources have finished loading.",
-            timeout=timeout
-        ).fulfill()
+        try:
+            # Wait for page to laod completely i.e. for document.readyState to become complete
+            EmptyPromise(
+                _is_document_ready,
+                "The document and all sub-resources have finished loading.",
+                timeout=timeout
+            ).fulfill()
+        except BrokenPromise:
+            LOGGER.warning(
+                u'document.readyState does not become complete for following url: {}'.format(self.url),
+                exc_info=True
+            )
+            # If document.readyState does not become complete after a specific time relax the
+            # condition and check for interactive state
+            EmptyPromise(
+                _is_document_interactive,
+                "The document is in interactive mode.",
+                timeout=timeout
+            ).fulfill()
 
         result = Promise(
             lambda: (self.is_browser_on_page(), self), "loaded page {!r}".format(self),


kashifch:
@feanil @estute @michaelyoungstrom 

Please review.

kashifch:
@feanil 

hmm that will pose some challenges.

The issue is that wait for page is not always called directly, mostly it is used by page.visit() in the back ground.

We can still implement a parameter passing in visit as well but what even this would impact only half the pages.

In e2e test repo we fetch pages from platform on the runtime before running tests, we only create pages in our repo which are not available in platform. So we can pass parameterized wait_for_page() or visit() in pages which are created in our repo but we won't be able to use these in the pages which are installed on the runtime from platform because these pages will still be using old wait_for_page(). If it were confined to only pages created in our repo I would have inherited the PageObject from bok-choy and modify it in e2e repo without any need to change bok-choy code.

I think the implementation I have done shouldn't have any negative impact, the relaxed condition is only called when the original wait for page has already failed, so in devstacks or sandboxes it will be virtually unused, it is only used when original condition fails and that mens that the test is already on the brink of failure.

I think if a couple of guys form Test Engineering can take a look and see if there could be any negative impact we might have a better idea.

Another thing we can do is to release an experimental version of bok-choy which we first only use in e2e repo and see if there are any adverse effect and then move it to platform as well.

These are just my thoughts, I am open to all suggestions

feanil:
@kashifch do pages in the platform implement their own `wait_for_page` logic?  I wouldn't expect that.  I would expect that since all pages inherit from this page object that adding the functionality here will make it available everywhere.  The concern I have with making your change the default is that it is misleading behavior since behavior.  We used to say that we are waiting for on-ready and now we wait on on-ready or on-interactive.  This is different behavior, it should be clear when this is in use.  I think you should add a `fallback_to_interactive` boolean parameter to `wait_for_page` and any core library functions that call it with a default of `False`.  So the current behavior works as is but the new behavior can be turned on in e2e tests.

feanil:
> [8:08 AM] Kashif Chaudhry: Hi Feanil, regarding the wait for page change, the obstacle I am talking about is that we use many pages from platform either directly in our tests or in inherited form. These pages are installed on the run time so I cannot make any change in them. These pages already have imported the original PageObject and uses the wait_for_page() in many functions. So if I add a boolean parameter to wait_for_page call I will only be able to use it in functions that I write myself, for functions already written in platform changes the wait_for_page() call will always be without parameter and hence will result in the same issue that we are trying to avoid unless I rewrite all these functions again in some wrappers
[8:09 AM] Kashif Chaudhry: for example
[8:09 AM] Kashif Chaudhry: 
    File uploaded: upload.png
    [https://s3.amazonaws.com/uploads.hipchat.com/26537/722973/0CWn9VMuRrzROkP/upload.png]
[8:10 AM] Kashif Chaudhry: the above is a page from platform which I am using in my tests but I cannot make any change a it is installed on runtime, notice the use of wait_for_page() in function and it is used in many of these
[8:11 AM] Kashif Chaudhry: I have no way to replace this call short of creating a new page and rewriting the function in it
[8:12 AM] Kashif Chaudhry: Please let me know if there is still any ambiguity and maybe we can setup a short call of 15 minute to sort this out

Given the complexity of wiring this all the way through all the existing objects, we need either a different way of flipping this switch or we need to not do it and accept that we are changing the default behavior.  Given the choices, I think it makes sense to pull in a few more opinions before we move forward with this.

feanil:
@jmbowman I'd love your thoughts on this.

jmbowman:
For context, jQuery switched its `ready()` method from triggering on `complete` to `interactive` in 2.2 ([PR](https://github.com/jquery/jquery/pull/2678), [release notes](http://blog.jquery.com/2016/01/08/jquery-2-2-and-1-12-released/)). We're currently using jQuery 2.2.4 in edx-platform.  So from that perspective, this really just compensates for a change in jQuery (which itself was motivated by gradual changes in the way browsers behave).

That said, it does concern me that we have pages which don't finish loading all assets within 30 seconds.  Can we get this to log which assets are still loading when the first promise fails?  That could help us improve the page performance and reduce test durations by identifying the outliers.

Regarding edx-platform compatibility: could you just create an edx-platform PR which installs bok-choy from this GitHub branch?  If all the tests pass with it, that would give us more confidence that this is a safe change to make.  If we choose to go ahead with it, we could then just update and repurpose that PR to upgrade bok-choy to the new release.

feanil:
@kashifch thoughts on what @jmbowman  is suggesting?  I think logging is a good addition. It would be good to know when and why we are falling back to interactive so that we could fix it.  That plus having a platform PR that tests the configuration as you currently have it seem like it would give me sufficient confidence that we wouldn't need to re-work this approach.

kashifch:
@feanil @jmbowman 

Thanks for the review and suggestions. I was a bit busy in other stuff. I will do it asap.

feanil:
Looks like the tests pass on the https://github.com/edx/edx-platform/pull/17659 PR.  @kashifch I think this will be good to go once we add the logging to figure out what we're still waiting for.  Does that sound right to you @jmbowman?

jmbowman:
No, this is definitely an improvement over the status quo.  It would be nice to be able to figure out which assets are failing to load in a timely manner, but that can be an enhancement for later.

feanil:
@kashifch can you merge this and do a new release of bok-choy?

kashifch:
Hi @feanil,

Sorry for delay, once thing came up in the last minute checkup and I have replaced the string concatenation with string format in the warning msg, it's a very minor change. Do I have to approve it as well or should I just go ahead with the merge

PR #203, opened 2018-03-15
Estute/disable a11y services

estute:


jmbowman:
We don't seem to need this anymore, tests are running fine without it.

PR #209, opened 2018-07-18
TE-2635 More logging on test failures, v0.8.0

jmbowman:
Logging improvements as noted in the CHANGELOG.  Hopefully this will help diagnose some of the recent flaky test failures in edx-platform.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ python:
  - '2.7'
  - '3.5'
  - '3.6'
- - 'pypy'
 sudo: false
 branches:
   only:
--- a/bok_choy/page_object.py
+++ b/bok_choy/page_object.py
@@ -328,6 +328,7 @@ class PageObject(object):
         try:
             self.browser.get(self.url)
         except (WebDriverException, socket.gaierror):
+            LOGGER.warning(u"Unexpected page load exception:", exc_info=True)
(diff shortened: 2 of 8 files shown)


bok_choy/page_object.py
```diff
@@ -372,6 +373,7 @@ def validate_url(cls, url):
                 # Valid URLs do not end with colons.
                 return False
         except ValueError:
+            LOGGER.warning(u"%s uses an invalid port", url)
```
estute: I wonder if we should add logging to all of the cases in which this function returns False. Where the function is called, it raises an exception if False that says  'Invalid' but is that helpful? What do you think? A lot of the time, the reasons for a URL are easy to spot, but this might be useful.

.travis.yml
```diff
@@ -3,7 +3,6 @@ python:
  - '2.7'
  - '3.5'
  - '3.6'
- - 'pypy'
```
estute: I support this

awaisdar001:
@jmbowman This is PR is in prod?

jmbowman:
Yes, it's already fully gone through the release process: https://github.com/edx/edx-platform/pull/18617 .  I forgot to upgrade it in edx-e2e-tests also, though; just opened a PR for that: https://github.com/edx/edx-e2e-tests/pull/300

PR #211, opened 2018-08-01
TE-2663 Support headless Chrome and Firefox

jmbowman:
* Added the `BOKCHOY_HEADLESS` environment variable to support use of headless Chrome or Firefox
* Don't try to save log types which aren't available for the browser in use
* Use the latest headless Firefox in Travis
* Updated dependency versions
* Updated a few tests to pass in Chrome also (one still fails due to needing a different screenshot baseline image)
* Prepared the next release


tox.ini
```diff
@@ -12,6 +12,9 @@ deps =
     -r{toxinidir}/requirements/test.txt
 passenv =
     DISPLAY
+    BOKCHOY_HEADLESS
+    MOZ_HEADLESS
```
estute: What is the purpose of this variable?

jmbowman: It's an alternative to the `-headless` command line argument to Firefox: https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-the-firefox-addon-in-headless-mode .  It's not technically needed, I just added it to make sure people used to activating headless mode that way don't get confused by it not working when running tox.

estute: Ok, cool. Thanks for the clarification

PR #202, opened 2018-03-08
#201 Allow extra desired capabilities

goodspark:
If this approach looks good to you all, I'll also push a commit to add some docs.

Ref: https://github.com/edx/bok-choy/issues/201

PR #213, opened 2018-11-13
bokchoy version 0.9.1

estute:
I want to push a new version to pypi containing this: https://github.com/edx/bok-choy/pull/202


jmbowman:
Just noticed the test failure, looks like a doctest needs to be updated.

PR #225, opened 2019-02-19
Update CI browser and deployment criteria

jmbowman:
* Use the latest stable releases of Firefox and geckodriver.
* Update the PyPI deployment criteria so it doesn't try to upload for both the `core` and `needle`Python 3.6 workers.


jmbowman:
@michaelyoungstrom Yeah, we still have this nightmare of a process for that: https://openedx.atlassian.net/wiki/spaces/TE/pages/814022796/How+to+upgrade+browser+versions

PR #226, opened 2019-02-26
update axe-core to 3.2.2

michaelyoungstrom:
Bump axe-core from 1.1 to 3.2.2 (latest version)

* `a11yCheck` has been removed from axe-core: [Link to changelog](https://github.com/dequelabs/axe-core/blob/develop/CHANGELOG.md#300-beta1-2018-02-19). I tried to just mimic how we called `axe.a11yCheck` with `axe.run` instead.
* A lot of our tests were failing due to the wrong expected failure counts. Seeing as how we're jumping 2 major releases, where they've added a lot of new rules, and based on some digging into the failures, this seems very reasonable to me. So I have adjusted the failure counts.
* Our unittest `test_limited_rule_tags` is no longer valid because the 1 rule that was failing in the past is no longer tagged with `wcag111`. I changed that test to represent a different tag instead. [changelog](https://github.com/dequelabs/axe-core/blob/develop/CHANGELOG.md#303-2018-06-04)


michaelyoungstrom:
@estute Yes, we will definitely need a new version to get this new axe-core version into platform

PR #229, opened 2019-05-01
Upgrade pip-tools and run make upgrade

michaelyoungstrom:
Similar to this PR: edx/devstack#401 , a manual upgrade of pip-tools to 3.6.1 was required to fix the auto dependency PRs

I also added the constraint of `urllib3<1.25` because `requests` supports `1.24` but `selenium` was trying to pull in the latest, which is `1.25.2` and was breaking `make upgrade`, and `sphinx<1.8.5` because it looks like pip-compile is failing to find its dependencies.


requirements/constraints.txt
```diff
@@ -13,3 +13,6 @@ more-itertools<6.0.0
 
 # Newer versions of pylint-plugin-utils drop Python 2.7 support
 pylint-plugin-utils<0.4
+
+# requests only supports 1.24 and selenium tries to pull in latest
```
jmbowman: Maybe link to the `requests` issue for resolving this? https://github.com/kennethreitz/requests/issues/5067

requirements/doc.txt
```diff
@@ -4,28 +4,14 @@
 #
 #    make upgrade
 #
-alabaster==0.7.12         # via sphinx
-babel==2.6.0              # via sphinx
```
jmbowman: All the sphinx dependencies disappeared from the file; that doesn't seem right, the `setup.py` on its master branch still lists these.

michaelyoungstrom: I'm stumped on this. I'm not sure why its removing them

jmbowman: Try adding `-v` to that pip-compile command to see if it gives more details?  When I run `make upgrade` in my local Python 3.6.8 environment, it seems to be updating correctly.

PR #294, opened 2020-11-30
BOM-2116 : Move to Github Actions

mzulqarnain1:
Relevant JIRA : https://openedx.atlassian.net/browse/BOM-2116


UsamaSadiq (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
...
+    name: Tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
+        python-version: ['3.8']
+        toxenv: ['core', 'needle', 'doc']
+
+    steps:
+    - uses: actions/checkout@v2
```
jmbowman, line 20: This should probably be `v1` to avoid problems with codecov, see https://openedx.atlassian.net/wiki/spaces/SRE/pages/2038792440/How+to+migrate+tests+from+TravisCI+to+Github+Actions for details.

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install pip
+        run: pip install -r requirements/pip.txt
+
+      - name: Install Dependencies
+        run: pip install setuptools wheel
```
jmbowman, line 24: I don't think this is needed after using `pip.txt` above.

